PR #56, opened 2024-10-30
Tweak spark test setup to tags and run tests appropriately

piyush-zlai:
## Summary
As of today our spark tests CI action isn't running the right set of Spark tests. The testOnly option seems to only include and not exclude tests. To get around this, I've set up a [SuiteMixin](https://www.scalatest.org/scaladoc/3.0.6/org/scalatest/SuiteMixin.html) which we can use to run the tests in a suite if there is a tag the sbt tests have been invoked with. Else we skip them all. 

This allows us to:
* Trigger `sbt test` or `sbt spark/test` and run all the tests barring the ones that include this suite mixin.
* Selectively run these tests using an incantation like: `sbt "spark/testOnly -- -n jointest"`. 

This allows us to run really long running tests like the Join / Fetcher / Mutations test separately in different CI jvms in parallel to keep our build times short. 

There's a couple of other alternative options we can pursue to wire up our tests:
* Trigger all Spark tests at once using "sbt spark/test" (this will probably bring our test runtime to ~1 hour)
* Set up per test [Tags](https://www.scalatest.org/scaladoc/3.0.6/org/scalatest/Tag.html) - we could do something like either set up individual tags for the JoinTests, MutationTests, FetcherTests OR just create a "Slow" test tag and mark the Join, Mutations and Fetcher tests to it. Seems like this requires the tags to be in Java but it's a viable option.

## Checklist
- [] Added Unit Tests
- [X] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

Verified that our other Spark tests run a bunch now (and now our CI takes ~30-40 mins thanks to that :-) ):
```
[info] All tests passed.
[info] Passed: Total 127, Failed 0, Errors 0, Passed 127
[success] Total time: 2040 s (34:00), completed Oct 30, 2024, 11:27:39 PM
```

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `TaggedFilterSuite` trait for selective test execution based on specified tags.
	- Enhanced Spark test execution commands for better manageability.

- **Refactor**
	- Transitioned multiple test classes from JUnit to ScalaTest, improving readability and consistency.
	- Updated test methods to utilize ScalaTest's syntax and structure.

- **Bug Fixes**
	- Improved test logic and assertions in the `FetcherTest`, `JoinTest`, and `MutationsTest` classes to ensure expected behavior.


nikhil-zlai (approved):
Fantastic PR!

PR #89, opened 2024-11-25
Frontend milestone

ken-zlai:
## Summary
This PR adds the following features:
- Wire up all charts to real data
- Distributions tab - view percentile charts for numerical data over time
- UX Improvements
    - Tooltip behavior near end of page
    - Chart series highlight on hover in legend, tooltip, title, etc
- Let user sort A-Z and Z-A
- Add drift metric toggle
- Consolidate chart controls

[Video demo and walkthrough](https://drive.google.com/file/d/1ZHaHd1yeEz0KPvQ-4FyN7xa4C3QEZuJj/view?usp=drive_link)

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `PercentileChart` component for visualizing percentile data.
  - Added a custom legend feature to the `EChart` component.
  - New `MetricTypeToggle` component allows users to switch between metric types.
  - Enhanced `ActionButtons` with sorting functionality.
  - New alert components for improved UI notifications.

- **Improvements**
  - Enhanced tooltip functionality in the `EChartTooltip` component.
  - Improved date range selection logic in the `DateRangeSelector` component.
  - Streamlined chart interaction logic in the `EChart` component.

- **Bug Fixes**
  - Fixed issues with mouse event handling in various chart components.

- **Chores**
  - Updated internal APIs to support new data structures and improve data fetching.

PR #120, opened 2024-12-10
sync the metric type toggle ui when url changes

ken-zlai:
## Summary
Stacked on #89. Quick change to make sure `MetricTypeToggle.svelte` stays synced with the url query params
[Demo](https://drive.google.com/file/d/1YNBxIi0fP-Qi-LVIw5FexwZQVEfeCWaS/view?usp=drive_link)

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


sean-zlai:
@ken-zlai I wonder if it's worth using [sveltekit-search-params](https://github.com/paoloricciuti/sveltekit-search-params) to streamline this a little.  `@next` version [supports](https://github.com/paoloricciuti/sveltekit-search-params/issues/80) Svelte 5 runes, although `@latest` with stores still work with Svelte 5 (used it on the take home [assessment](https://github.com/techniq/zipline-take-home/blob/main/src/routes/%2Bpage.svelte#L23)).

I am hopeful they will add "first-party" support to SvelteKit as part of the [Advent of Svelte](https://svelte.dev/blog/advent-of-svelte), but haven't seen any indication yet (but 14 days to go :) ).

Also hopeful they include strong-typed routes similar to [vite-plugin-kit-routes](https://www.kitql.dev/docs/tools/06_vite-plugin-kit-routes) (and similar to [TanStack Router](https://tanstack.com/router/latest)).

ken-zlai:
@sean-zlai This is a good recommendation, and I did want to use [sveltekit-search-params](https://github.com/paoloricciuti/sveltekit-search-params). However, I don't think the library supports server side code. Meaning, we would need two approaches to manage query paramaters.
- One used in client side code
- Another used in the server side code

From my perspective, having one function/source of truth is preferable

sean-zlai (approved):
We talked offline and there might still be an opportunity to use something like `sveltekit-search-params` (or something first party in SvelteKit if added) to streamline syncing query string state, but that can be investigated in a downstream PR.

PR #134, opened 2024-12-17
fix: since search only supports joins, show them all and only enable …

ken-zlai:
## Summary
Right now our search endpoint only returns joins. I change the code to account for this by showing all results as a join. I have also only made `risk.user_transactions.txn_join` clickable, since it is the only one to have data.

In the future, we should add an `entityType` to search results to categorize them on the frontend.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
	- Enhanced rendering of search results in the navigation bar with improved handling of entity properties.
  
- **Bug Fixes**
	- Fixed conditional display logic for command items based on entity names. 

- **Refactor**
	- Streamlined the selection logic and rendering process for search results by replacing the `model` variable with `entity`.


sean-zlai (approved):
LGTM, and thanks for fixing this.  I ran into it while working on the switch to unplugin-icons and regression testing but didn't look into it since it was also an issue on AWS/staging.

PR #177, opened 2025-01-07
GroupByUploader in Driver should use the top level TableUtils

david-zlai:
## Summary

- GroupByUploader in Driver should use the top level TableUtils. This is necessary because in the top level spark session of Driver.scala, we add additional configs. If we don't pass it in, GBU will create a new spark session that doesn't get the good stuff from Driver

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Improvements**
	- Enhanced `GroupByUpload` functionality by adding optional table utilities parameter to the `run` method
	- Expanded method capabilities for more flexible configuration and processing


tchow-zlai (approved):
LGTM

PR #190, opened 2025-01-09
Remove Thrift-generated files from Git repository

ken-zlai:
## Summary
This PR removes all Thrift-generated files from the Git repository. These files are already part of the build process, so they will still be generated and usable locally

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Chores**
	- Updated `.gitignore` to exclude `/frontend/src/lib/types/codegen`
	- Removed multiple TypeScript type definition files related to API, common, observability, and orchestration types

- **Refactor**
	- Deleted generated type definition and implementation files, likely as part of a code cleanup or restructuring effort


sean-zlai (approved):
While I think there might still be some benefit to checking in the generated types, overall generating these types "on demand" will work.

A few future ideas we might want to consider:

Once we start importing / referencing these types (ex. `import type { SomeType } from '$lib/types/codegen'` we'll need to also call `npm run codegen` before `lint`, `check`, and all the `test` [scripts](https://github.com/zipline-ai/chronon/blob/main/frontend/package.json#L11-L20).

Since this can get really repetitious to do it for each script in `package.json`, it might be better to add an explicit `Generate thrift types` step in our CI [workflow](https://github.com/zipline-ai/chronon/blob/main/.github/workflows/test_frontend.yaml#L46) (after `npm ci` and before `check`/`lint`/etc).

Locally, re-running `npm run dev`, or explicitly `npm run codegen` after pulling API changes from `main` should be sufficient to keeping them in sync.  We could add a note to the frontend `README.md` if needed.

It might also be useful to run the `Test Frontend` workflow after any [api/thrift/*](https://github.com/zipline-ai/chronon/tree/main/api/thrift) changes to monitor breaking changes that need addressed.

PR #244, opened 2025-01-17
Route and component restructuring (includes performance improvements)

sean-zlai:
## Summary

- Removed directory nesting for components (except shadcn-svelte components in `ui`)
- Rework page layout structure and data loading strategy
  - Removes unnecessary data loading (ex. data distributions)
  - Improves performance by no longer rendering other/offscreen tab content
  - Provides better page component organization (and smaller files) via use of layouts (including nested)
  - Data loading can be handled more granularly and consistently via `load()` function.  This also enables browser caching via `setHeaders({ 'cache-control': ... })` ([example](https://github.com/techniq/strava-analysis/blob/main/src/routes/analysis/%2Blayout.server.ts#L73C3-L75C6))
  - Persist tab state for drift/distributions as route params

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
	- Simplified import paths across multiple frontend components
	- Removed redundant subdirectories in component imports

- **New Features**
	- Added new layout for joins detail page with tab navigation
	- Introduced new observability navigation tabs
	- Created new ModelTable component for displaying model details
	- Added new distributions page for feature distributions
	- Implemented zoom functionality in distribution charts
	- Added placeholder content for lineage page

- **Changes**
	- Updated page routing for joins detail pages
	- Modified page load and redirect behaviors
	- Restructured observability and job tracker views
	- Added loading state handling in distribution components
	- Implemented redirects for specific observability routes

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209182128830925

PR #251, opened 2025-01-20
Migrate ECharts usage to LayerChart

sean-zlai:
## Summary

Some of the benefits using LayerChart over echarts/uplot

- Broad support of chart types (cartesian, polar, hierarchy, graph, force, geo)
- Simplicity in setup and customization (composable chart components)
- Responsive charts, both for viewport/container, and also theming (light/dark, etc)
- Flexibility in design/stying (CSS variables, classes, color scales, etc) including transitions
- Ability to opt into canvas or svg rendering context as the use case requires. LayerChart's canvas support also has CSS variable/styling support as well (which is unique as far as I'm aware). Html layers are also available, which are great for multiline text (with truncation).

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Introduced new interactive charts, including `FeaturesLineChart` and `PercentileLineChart`, enhancing data visualization with detailed tooltips.
- **Refactor**
  - Replaced legacy ECharts components with LayerChart components, streamlining chart interactions and state management.
- **Chores**
  - Updated dependency configurations and Tailwind CSS settings for improved styling and performance.
  - Removed unused ECharts-related components and functions to simplify the codebase.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209163154826936


sean-zlai:
Creating this PR to replace #214 as the merge conflicts from #244 were easier to handle from scratch

nikhil-zlai (commented):
@ken-zlai and I talked about changing the endpoints to no-longer produce time-points and directly consume double arrays and timestamp arrays  - as defined within TileSummarySeries and TileDriftSeries thrift objects (using the thrift router work that is merged recently)

It might save you effort to do this migration on top of what Ken was planning to do. Or you could even pick that task up. Happy to walk you through what needs to be done on that task as well.:-)

sean-zlai:
Thanks for the heads up @nikhil-zlai.  Might be good for us to sync up to make sure I understand everything.  I see these thrift generated interfaces in my local codegen'd files.

image
 
Is the driver for this change to reduce the amount of over-the-wire transfer size (and less transforming on the backend)?

nikhil-zlai:
> Might be good for us to sync up to make sure I understand everything. 

Lets do it! Please throw a meeting on my cal any day this week after 11 PST.

> Is the driver for this change to reduce the amount of over-the-wire transfer size (and less transforming on the backend)?

That's right. 

Sending separate arrays of floats and timestamps is optimal, specially when there are many series in a single chart - we don't need to duplicate timestamps for each series / line, and we don't need to duplicate the label string.  Basically this positions the front-end to deal with as few strings as possible.

So, one additional benefit is also to reduce the amount of transformation needed in the front-end library. 

For example, echarts/uplot takes separate float arrays per series and timestamp array.

sean-zlai:
At the moment LayerChart does not support separating out per-dimension data (but does support data per series either on a common data structure (property per series) or data split per series.  Supporting separating dimension data will take some thought, but we can transform the data client side for now (and still have the over the network benefits).

As I type and think about this, this might be possible to handle via the dimension accessors as it's just needs look up based on index position from 2 different arrays (although I'm not sure if the index position is passed in all contexts).

Oh well, something to think about, and transforming client side is still possible (just more intensive on the client).

nikhil-zlai:
how do you feel about doing some simple benchmarking prior to our meeting?

we can try to use arrays per label like in echart / uplot, vs map per tick like in layer chart in a simple benchmark.

for baseline: https://leeoniya.github.io/uPlot/bench/uPlot.html

I am very curious to see the delta in perf.

sean-zlai:
> how do you feel about doing some simple benchmarking prior to our meeting?
> 
> we can try to use arrays per label like in echart / uplot, vs map per tick like in layer chart in a simple benchmark.
> 
> for baseline: https://leeoniya.github.io/uPlot/bench/uPlot.html
> 
> I am very curious to see the delta in perf.

I'll try to setup some benchmarks before our meeting, but TBH I wouldn't be surprised if LayerChart isn't a few orders slower than uplot based on their [published benchmarks](https://github.com/leeoniya/uPlot?tab=readme-ov-file#performance) since that is their primary focus.

image

I'll start by setting up a LayerChart example using [data.json](https://leeoniya.github.io/uPlot/bench/data.json) from their benchmark and see what kind of limits I encounter.

Some of the benefits using LayerChart over echarts/uplot 

- Broad support of chart types (cartesian, polar, hierarchy, graph, force, geo)
- Simplicity in setup and customization (composable chart components)
- Responsive charts, both for viewport/container, and also theming (light/dark, etc)
- Flexibility in design/stying (CSS variables, classes, color scales, etc) including transitions
- Ability to opt into canvas or svg rendering context as the use case requires.  LayerChart's canvas support also has CSS variable/styling support as well (which is unique as far as I'm aware).  Html layers are also available, which are great for multiline text (with truncation).

Here's a quick recording of this PR using LayerChart (using SVG render context ATM but might switch to canvas in the future with a simple `renderContext="canvas"` prop)

https://github.com/user-attachments/assets/92e4a06f-88ad-4a51-ada1-5a987b5f6551

Performance wise "feels" very similar to the current echarts solution, but I've also also improved the tooltip handling (tracking data with smart container positioning) and replaced the null data ratios with pie charts (which wouldn't be possible with uplot).  There is a lot of performance tuning that can be improved on, especially on pages with multiple charts (only rendering based on viewport visibility, etc).

ken-zlai:
Chiming in here - I was going to start this thrift migration today ([ticket](https://app.asana.com/0/1208277377735902/1209184028451140/f)). However, given this migration and our chat tomorrow, I can pick up a few other tasks in the meantime to keep me busy.

We can decide tomorrow on who it makes the most sense to convert existing api's -> thrift.

ken-zlai (approved):
@sean-zlai This is really excellent work here, code largely looks good. Charts are snappy and the net-negative code is awesome!

A couple of things, mainly UI related:
1. I noticed that things like the tooltip and legend are slightly different than the figma/previous version (mainly padding and sizing). Curious if we plan on matching these in the future.
2. One improvement would be to show a [cursor: pointer](https://tailwindcss.com/docs/cursor) when they are hovering over a clickable point
pointer
3. The legend previously had an [expand/hide functionality](https://github.com/zipline-ai/chronon/pull/251/files#diff-0675b3f1a79cde74860a954421cbf66f1c4e4ee61b022cdeff8837bc1a4f8cd6L116) if there was overflow. Meaning, if there were too many series you could click a button to show/hide them all. Now I notice that it scrolls left and right. I'm not picky on the treatment, but think that the expand functionality is a bit more intuitive. However, I remember it being a bit tricky to implement so if reusing the code I had is not easy then lets not do it.

Let me know if we plan on fixing here/in the future. I'm happy to create a few tickets if we want to.

sean-zlai:
> @sean-zlai This is really excellent work here, code largely looks good. Charts are snappy and the net-negative code is awesome!

Thanks, and great feedback!  See responses below

> 
> A couple of things, mainly UI related:
> 
> 1. I noticed that things like the tooltip and legend are slightly different than the figma/previous version (mainly padding and sizing). Curious if we plan on matching these in the future.

I had tighten it up a little, but we can definitely align to the design.  It's mostly tweaking padding/margin classes [here](https://github.com/zipline-ai/chronon/pull/251/files#diff-d04278c87a5269ac80683fb382b1e8d6c5157c81fdfb458cc8fd8b1db494f440R26-R44).  I'll take a look and adjust.

> 2. One improvement would be to show a [cursor: pointer](https://tailwindcss.com/docs/cursor) when they are hovering over a clickable point
>    pointer

Very good feedback.  If it was using `renderContext="svg"` (default) it would be a simple:
```svelte
<BarChart props={{ highlight: { points: { class: 'cursor-pointer' } } }}>
```
but since we're using `<BarChart renderContext="canvas">` for better performance, it will be a little trickier.  With the recent [introduction](https://bsky.app/profile/techniq.dev/post/3lgzyqjqndc2v) of canvas pointer events I can to wire up a `onpointerenter` / `onpointerleave` that conditionally sets `cursor-pointer` on the `<canvas>` container.  Let me look into that (might even do it in LayerChart as when a `Highlight` point has an click, it makes sense to always set it I think.

> 3. The legend previously had an [expand/hide functionality](https://github.com/zipline-ai/chronon/pull/251/files#diff-0675b3f1a79cde74860a954421cbf66f1c4e4ee61b022cdeff8837bc1a4f8cd6L116) if there was overflow. Meaning, if there were too many series you could click a button to show/hide them all. Now I notice that it scrolls left and right. I'm not picky on the treatment, but think that the expand functionality is a bit more intuitive. However, I remember it being a bit tricky to implement so if reusing the code I had is not easy then lets not do it.
> 
> Let me know if we plan on fixing here/in the future. I'm happy to create a few tickets if we want to.

Yeah, I looked at this for a bit and the biggest challenge is the `Legend` is within the `<Chart>` and absolutely positioned to the bottom of the container.  Expanding it with this positioning makes it expand upward into the chart, which isn't optimal.  I could look at resizing the container when this expansion happens, but since there's no guarantee "adding a second line" will accommodate all options, the nuance was going to take some thought.

I also considered implementing a button that scrolls the overflow for you (kind of similar to a carousel), similar to the Skeleton Multi Column [example](https://www.skeleton.dev/elements/scroll-containers) or MUI's scrollable tabs [example](https://mui.com/material-ui/react-tabs/#scrollable-tabs).  This would allow the legend to remain a single line, but with a nice way to paginate through all the options without requiring horizontal swiping, which isn't convenient on a traditional mouse (trackpad and magic mouse aren't bad though).

I could also look into exposing more chart state externally (already do for tooltip and brush) but it would take some thought.

I'll knock out #1 and #2 and let me know if #3 is worth waiting before we merge.

Thanks for the review!

ken-zlai:
@sean-zlai Sounds like a plan to get #1 and #2. As for #3, I'm open to a simple treatment - the MUI's scrollable tabs is generally nice since you can see when there is overflow/click thru if needed.

sean-zlai:
@ken-zlai #1 and #2 are done.  The tooltip items are not 100% matching, but much closer (removed vertical gap between items, gave each item more height via padding, and increased gap between label and value).

#3 will take some time.  Would you be OK if we merge as is and circle back after landing #318?  That PR is really close once I figure out the backend `NullPointerException` for some features and do some more cleanup.

ken-zlai:
@sean-zlai Sounds good to me! Here is a [ticket for #3 ](https://app.asana.com/0/1209162816962686/1209343349099312)

PR #332, opened 2025-02-06
fix: ooms while reading large columns

nikhil-zlai:
## Summary
the default column reader batch size is 4096 - reads that many rows into memory buffer at once.
that causes ooms on large columns, for catalyst we only need to read one row at a time. for interactive we set the limit to 16.

tested on etsy data.
 
## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
	- Enhanced data processing performance by adding an optimized configuration for reading Parquet files in Spark sessions.


piyush-zlai (approved):
Maybe also add a comment on the rationale for future folks?

nikhil-zlai:
> Maybe also add a comment on the rationale for future folks?

done

tchow-zlai (approved):
LGTM

PR #439, opened 2025-02-26
Implement EMR submitter

david-zlai:
## Summary

^^^

Notes:

Ideally, we wanted to just spin up a single long running cluster and submit spark jobs adhoc whenever we want but that setup looks to only be supported by EMR on EKS. We're currently EMR on EC2. 

For EMR on EC2, it uses the EMR [RunJobFlow API](https://docs.aws.amazon.com/emr/latest/APIReference/API_RunJobFlow.html). It will create a cluster and run a list of spark jobs (or steps) and then terminate the cluster (or not if we set it to). Thus, one cluster, one JobFlow. And a JobFlow can only [have 256 steps](https://docs.aws.amazon.com/emr/latest/APIReference/API_AddJobFlowSteps.html) or 256 spark jobs in that JobFlow. Thus, we have a tradeoff where we can't have a true long running cluster with EMR on EC2 since once a cluster hits the 256 steps limit, the EMR api won't accept any more steps to that JobFlow. _(one hack though to get around this is to submit spark jobs via ssh'ing on the master node and not through EMR api)_

For our case we only have one spark job so thus only one "step" in an EMR Job. 

This PR will call the RunJobFlow API for a single Chronon spark job or step, and create the cluster for just that single spark job. If the spark job or step succeeds or fails, we'll keep the cluster alive and terminate it when it becomes idle after some time. 

Another important note is that we are submitting jobs to EMR via the [command-runner.jar](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-commandrunner.html) approach. This looks a little different than Dataproc where with EMR we are using a jar they provide that is just a runner to ultimately spark-submit our cloud_aws jar and call **Driver**.  Tried to keep EMR's submission consistent with Dataproc but ran into issues. Explained more [here](https://docs.google.com/document/d/1bQnTOK8P3Spga2sm9Y1eacQY8H3uwv4asrCnuV_jOws/edit?tab=t.0#heading=h.php300tv32f8)

Finally, EMR doesn't offer the same Dataproc `files` approach where in Dataproc you can set GCS files that will be copied onto the cluster instances. Instead, we utilize bootstrap actions in EMR to do this and code up a simple shell script that takes in s3 files and downloads them as part of bootstrap startup of the emr cluster

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Introduced AWS EMR job submission capabilities, allowing users to submit, monitor, and cancel jobs.
- **Enhancements**
  - Improved command-line argument handling and configuration for smoother job submission workflows.
  - Added support for the latest AWS EMR SDK in both main and testing environments.
- **Tests**
  - Added comprehensive unit tests to ensure reliable integration and performance.
- **Chores**
  - Updated build configurations and dependency management to include the latest AWS EMR SDK alongside improvements in related libraries and artifacts.


david-zlai:
i'm going to push some changes to instead submit jobs to an existing cluster instead of spinning up a new cluster.

tchow-zlai (approved):

PR #444, opened 2025-02-26
table typeahead

ken-zlai:
## Summary
Implemented fuzzy search for EntityTable using [Fuse.js](https://www.fusejs.io/) with default match settings.

https://github.com/user-attachments/assets/edaa3a89-1a15-49b9-b8ec-fb2e768f3c04

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
	- Introduced a search capability within the table view, allowing users to quickly filter items based on relevant attributes.
	- The search interface now automatically resets after navigation, providing a cleaner and more intuitive user experience.


ken-zlai:
@sean-zlai This is ready for another look!

sean-zlai:
@ken-zlai getting some odd behavior as I type into the typeahead (might be how `fuse.js` works, but feels very odd to jump between more/less results).

https://github.com/user-attachments/assets/7e9446e6-d7b0-47ea-a5ce-668f3b4407f0

I don't recall seeing this earlier, but maybe I overlooked it.  Might be fine to just use a simple [String.prototype.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)`

ken-zlai:
Fair enough, yeah I'll just remove fuse. It seemed nice but I also felt like some of its matching logic was not the greatest

ken-zlai:
**@sean-zlai Done. Just a heads-up:** This approach requires exact substrings to match. For example, searching for `"risk user"` won’t return `"risk.user_transactions.txn_join"`, but `"risk.user"` will. If we need more flexible matching (e.g., handling spaces as separators), we’d have to adjust the logic.

sean-zlai:
> Fair enough, yeah I'll just remove fuse. It seemed nice but I also felt like some of its matching logic was not the greatest

Same.  I'm sure we could tweak a bunch of it's options, and maybe something I'll look at down the road if we need more robust searching.

Thanks for the quick turnaround

sean-zlai (approved):
LGTM, thanks!

PR #453, opened 2025-02-28
Update svelte-inspect-value and use new css/theme variables

sean-zlai:
## Summary

Using the new [recently](https://github.com/ampled/svelte-inspect-value/issues/23#issuecomment-2690391648) released extended css/theme variables which allows to mute the hierarchy lines but not the caret.

This also affected some of the default colors, so I added more theme colors to (mostly) align to the previous colors (without these the values were mostly white).

Before | After
--- | ---
image | image

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


ken-zlai (approved):
much cleaner!

PR #456, opened 2025-02-28
Fix chart prop merging

sean-zlai:
## Summary

I noticed after we added the `cursor-crosshair` canvas prop the other props like radius size were no longer being applied.  This fixes it by deeply merging all the different prop sources together

Before | After
--- | ---
image | image

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


ken-zlai (approved):
great catch! works for me

PR #469, opened 2025-03-05
Improve PercentileLineChart perf (and thus Summary page) by over 20x

sean-zlai:
## Summary

By extracting `series` from the markup (which made it reactive) and intentionally NOT using `$derived`, rendering performance of `PercentileLineChart` improved by over 20x (`>1100ms` down to `<50ms`).

This makes navigating to the Summary tab MUCH better (before we would sometimes hit the browser "Are you sure you want to wait" prompts).

`$state.opaque()` runes was being [considered](https://github.com/sveltejs/svelte/pull/14639) that could have helped here, but it was decided to not be the right abstraction.  (I'm waiting on a solution to fix LayerChart's force simulations that have a perf regression in Svelte 5 due to granular reactivity / proxy creation)

## Before

https://github.com/user-attachments/assets/9e93f4eb-5247-4bff-a56d-413564e50022

## After

https://github.com/user-attachments/assets/8dbe075d-0479-4d5e-a3f8-45fb1d350402

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Optimized the chart component's data handling to improve performance and overall clarity.
  - Adjusted the internal structure by moving data definitions outside inline usage and refining type specifications for better consistency.


ken-zlai (approved):
this is insane - amazing! loading the summary page is super snappy now. great job!

PR #475, opened 2025-03-06
fix: adjust classpath so flink side can work

tchow-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Chores**
  - Removed legacy resource management components.
  - Updated dependency configurations by excluding outdated libraries and incorporating essential new ones, including new dependencies for Hadoop and Jetty.
  - Upgraded several third-party libraries to enhance compatibility and maintainability.


piyush-zlai (approved):

piyush-zlai (changes requested):
Actually it looks like we potentially still need to iterate here. I'm consistently seeing the Flink UI being partially broken. I see the job running (via logs, checkpoints and kv store write metrics) but we're not able to look at the DAG view and explore. The UI seems to think that - "Job is not running yet". We do need the UI functional for a lot of debugging / diagnostic stuff so we should resolve..

piyush-zlai (approved):
Flink wise the current version now looks good. We had to add commons-text to exclusions as the execution graph wasn't being loaded (and hence that api was returning no rows) - `2025-03-06 18:25:46,729 WARN  org.apache.flink.runtime.jobmaster.JobMaster                 [] - Cannot create JSON plan for job
java.lang.NoSuchMethodError: 'org.apache.commons.lang3.Range org.apache.commons.lang3.Range.of(java.lang.Comparable, java.lang.Comparable)'`

PR #487, opened 2025-03-07
fix: bootstrap unit tests should be deterministic

tchow-zlai:
## Summary

- In certain tests we will generate some dummy data with random values. We proceed to use this data to do correctness checks. However, since dataframes are lazily evaluated, whenever we reference the value of the dataframe, the random generation kicks in. Thus, we sometimes get incorrect unit test results based on the data that gets generated at the time of referencing the dataframe.  The fix is to first save the dataframe to a table, and load it back immediately to preserve the values that were generated the first time.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Standardized the approach to loading table data in test cases for improved consistency and maintainability.
  - Updated variable names to enhance clarity regarding the data being handled.
  
- **Tests**
  - Improved test logic with clearer variable naming and a refined process to ensure tests reflect the most recent data state after saving.
  - Minor formatting adjustment in the test case for better readability.


nikhil-zlai (approved):
great find!

PR #539, opened 2025-03-26
chore: remove caching and repartitioning feature

tchow-zlai:
## Summary

- Remove the repartitioning and caching. We'll let AQE handle this for us, and we can just tune those params which are open. 

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Streamlined data persistence operations by simplifying how data is saved and partitions are handled, reducing unnecessary complexity.
  
- **Chore**
  - Consolidated dependency management and improved overall code organization to promote clarity and maintainability.
  
- **Bug Fixes**
  - Removed the "column size estimator" test case, eliminating associated logic and assertions.

These changes aim to enhance the efficiency and reliability of data processing without altering user-facing functionality.


piyush-zlai (approved):
Changes lgtm - A q, do we feel like we might want the escape hatch of being able to sidestep aqe and use repart and write?

tchow-zlai:
> Changes lgtm - A q, do we feel like we might want the escape hatch of being able to sidestep aqe and use repart and write?

Hard to say what scenario that might fall under, but I do think the repartitioning would be harder for users to reason about (and to control). Even if we come across a job that AQE handles poorly, In that situation I feel like we'd have to tune specific to the job rather than rely on the current general repartitioning algorithm.

tchow-zlai:
cc @nikhil-zlai

PR #736, opened 2025-05-05
chore: Use table caching for bigquery

tchow-zlai:
## Summary

- It is important to include the partition filters as part of the table cache key. Otherwise, this can actually return incorrect results since we push the partition filters down into the table scan, so it is not visible to the cache. 
- Also wrap BigQuery native table scans in the table cache

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Improved table caching to account for both table name and partition filters, leading to more accurate data retrieval.
  - Simplified and unified method signatures for partition handling, with some parameters now having default values.
  - Removed redundant method overrides for partition operations to streamline codebase and reduce duplication.

- **Chores**
  - Updated import statements for consistency and maintainability.


spark/src/main/scala/ai/chronon/spark/catalog/Format.scala
```diff
@@ -30,18 +29,22 @@ trait Format {
   def table(tableName: String, partitionFilters: String, cacheDf: Boolean = false)(implicit
       sparkSession: SparkSession): DataFrame = {
 
-    val df = if (cacheDf) {
-      TableCache.get(tableName)
+    if (cacheDf) {
```
nikhil-zlai, line 32: i am just worried that this will hurt performance of beacons. 

can you do a quick check on the join with and without?

tchow-zlai: I agree that this will hurt the performance. No worries, with your change to include the partition column in the partition spec I can try to push that all the way down to give us some better options around table scanning. Can discard this PR.

PR #776, opened 2025-05-15
replace partition spec with column

varant-zlai:
## Summary

replace partition spec with column -> partiton spec

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Simplified partition specification handling across planners and utilities by removing the custom partition spec wrapper and standardizing on a single partition spec type.
  - Updated related methods and class constructors to use the new partition spec approach, streamlining partition metadata access.
  - Removed unused fields and imports related to the old partition spec wrapper.


nikhil-zlai (changes requested):
why not use the partition column instead of hardcoded dt everywhere?

nikhil-zlai (approved):

PR #793, opened 2025-05-20
perf: shuffle-free temporal join

nikhil-zlai:
## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

- **New Features**
  - Introduced new utility functions for efficient time-series aggregation, including sawtooth pattern point-in-time aggregation.
  - Added `UnionJoin` for unioning and normalizing DataFrames with schema alignment, filtering, and timestamp-based sorting.
  - Implemented `AggregationInfo` to encapsulate aggregation logic, schemas, and Spark UDF integration.
  - Added a skew-free partition filling mode for join backfill processing.
  - Created a shell script for executing performance tests with heap dump and memory profiling.
  - Added a method to prepare unified input DataFrames for group-by operations with join source replacement and key filtering.

- **Bug Fixes**
  - Improved filtering to exclude rows with null timestamps or keys during join and aggregation processes.

- **Tests**
  - Added extensive test suites for `unionJoin`, `SawtoothUdf.sawtoothAggregate`, and performance validation with large datasets.
  - Included correctness assertions for aggregation results, sorting order, filtering, and handling of empty inputs.
  - Added tests verifying join execution and output correctness for temporal event scenarios.

- **Documentation**
  - Enhanced code readability through formatting adjustments across multiple components.


spark/src/main/scala/ai/chronon/spark/join/UnionJoin.scala
```diff
@@ -0,0 +1,179 @@
...
+          left_data_array_unsorted,
+          (left, right) -> CAST(left.$leftTimeCol - right.$leftTimeCol AS INT)
+        )
+      """).as("left_data_array") :+
+          F.expr(s"""
+        array_sort(
+          right_data_array_unsorted,
+          (left, right) -> CAST(left.$leftTimeCol - right.$leftTimeCol AS INT)
+        )
+      """).as("right_data_array"): _*
```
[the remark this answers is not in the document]
nikhil-zlai: spark sql requires it to be int

scripts/perf/run_unionjoin_with_manual_dump.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
+# UnionJoinTest with Manual Heap Dump Generation
+
+cd /Users/nsimha/repos/chronon
```
david-zlai, line 5: Maybe we add this file in platform instead of chronon?

spark/src/main/scala/ai/chronon/spark/Driver.scala
```diff
@@ -275,6 +276,30 @@ object Driver {
...
+        val startPartition = args.startPartitionOverride.toOption.getOrElse(args.joinConf.left.query.startPartition)
+        val endPartition = args.endDate()
+
+        val joinName = args.joinConf.metaData.name
+        val stepDays = args.stepDays.toOption.getOrElse(1)
+
+        logger.info(
+          s"Filling partitions for join:$joinName, partitions:[$startPartition, $endPartition], steps:$stepDays")
+
+        val partitionRange = PartitionRange(startPartition, endPartition)(tableUtils.partitionSpec)
```
david-zlai, line 291: We should be careful using `tableUtils.partitionSpec`  because it defaults to the spark.chronon.partition.format [here](https://github.com/zipline-ai/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/catalog/TableUtils.scala#L51-L53). 

Defaulting to do that led to this [bug](https://github.com/zipline-ai/chronon/pull/783/files#diff-792dc9c095a638390fecab35c2ffe8449e38576343a5ebba9993f93c47b1429aR176) because users can technically define a custom partitionSpec at the Source level. 

we could do something like:

```
        val partitionRange = PartitionRange(startPartition, endPartition)(args.joinConf.left.partitionSpec)

```

david-zlai: do we know if all of their tables follow a consistent partition spec now? I remember we did ask them to staging query that one additional table. so maybe we're good here now

varant-zlai (approved):

PR #834, opened 2025-06-04
Include offset in airflow dependency name

david-zlai:
## Summary

This also helps prevent collisions when we have multiple airflow dependencies to the same table with different offsets. We don't want them to overwrite each other

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

No user-facing changes in this release.


api/python/ai/chronon/airflow_helpers.py
```diff
@@ -43,8 +43,12 @@ def create_airflow_dependency(table, partition_column, additional_partitions=Non
     if additional_partitions:
         additional_partitions_str = "/" + "/".join(additional_partitions)
 
+    wf_name = f"wf_{utils.sanitize(table)}"
+    if offset != 0:
```
tchow-zlai, line 47: let's just always include the offset (even if 0) to be consistent.

tchow-zlai (approved):

PR #908, opened 2025-06-27
Throw specific exception for when cluster is not in healthy state.

david-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unhealthy Dataproc clusters by explicitly notifying users when an existing cluster is not in a healthy state before attempting to create a new one.  
  * Added validation to prevent creation attempts when an existing cluster is detected as unhealthy.

### DIFF
--- a/cloud_gcp/src/main/scala/ai/chronon/integrations/cloud_gcp/DataprocSubmitter.scala
+++ b/cloud_gcp/src/main/scala/ai/chronon/integrations/cloud_gcp/DataprocSubmitter.scala
@@ -525,7 +525,6 @@ object DataprocSubmitter {
     if (clusterName != "") {
       try {
         val cluster = dataprocClient.getCluster(projectId, region, clusterName)
-
         if (cluster != null) {
           if (
             Set(ClusterStatus.State.RUNNING, ClusterStatus.State.UPDATING, ClusterStatus.State.CREATING).contains(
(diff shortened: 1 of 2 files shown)


cloud_gcp/src/main/scala/ai/chronon/integrations/cloud_gcp/DataprocSubmitter.scala
```diff
@@ -524,13 +525,19 @@ object DataprocSubmitter {
...
+          if (
+            Set(ClusterStatus.State.RUNNING, ClusterStatus.State.UPDATING, ClusterStatus.State.CREATING).contains(
+              cluster.getStatus.getState)
+          ) {
+            println(s"Dataproc cluster $clusterName already exists and is healthy.")
+            clusterName
+          } else {
+            throw new Exception(
+              s"Dataproc cluster $clusterName exists but is not in a healthy state: ${cluster.getStatus.getState}. " +
+                s"Attempting to create a new cluster with the provided config.")
```
tchow-zlai, line 539: Doe throwing a generic `Exception` actually cause us to try to create a new cluster?  it looks like down below we do that for `ApiException` only.

david-zlai: it shouldn't. Throwing this generic exception should exit the entire code

tchow-zlai (approved):
forgot about this one but lmk if we need to update it based on some of @chewy-zlai 's recent changes.

PR #919, opened 2025-07-01
Fix: Disable Bazel Remote Cache

chewy-zlai:
## Summary

No longer use bazel remote cache as it requires using a secret which isn't available to external collaborators.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Chores**
  * Simplified continuous integration workflows by removing remote caching and credential configurations from build and test steps, streamlining the process without affecting user-facing features.


piyush-zlai (approved):
Could just turn off the bazel cache - not sure if its giving us a lot perf wise

PR #944, opened 2025-07-08
Feat: Use GCP Credentials in Zipline Hub

chewy-zlai:
## Summary

When the base_url of orchestration ends with ".app", pass in gcp credentials for zipline hub calls.
Also switch to using post requests for the diff API.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Enhanced authentication for API calls when connecting to ".app" endpoints, providing secure access using Google Cloud credentials.
* **Improvements**
  * All API requests now use JSON content-type headers and include authorization tokens when available.
  * Improved consistency in API request methods and headers for better reliability.


api/python/ai/chronon/repo/zipline_hub.py
```diff
@@ -1,22 +1,32 @@
...
 
     def call_diff_api(self, names_to_hashes: dict[str, str]) -> Optional[list[str]]:
         url = f"{self.base_url}/upload/v1/diff"
 
         diff_request = {
             'namesToHashes': names_to_hashes
         }
+        headers = {'Content-Type': 'application/json'}
+        if hasattr(self, 'id_token'):
+            headers['Authorization'] = f'Bearer {self.id_token}'
```
piyush-zlai, line 27: it might be worrisome to have this auth token sent over as a header if we're using http - someone sniffing the network traffic can get access to this token and hence gcloud resources for the ~hour that the token is valid. 
Might be ok as a short-term stopgap but we should try and track and swap out.

piyush-zlai (approved):
One item to call out w.r.t. http / https. We can try and address either upfront (so starting with hub and backwards) or as a follow up.

PR #956, opened 2025-07-10
Add a node runner for upload to kv and join metadata upload nodes

piyush-zlai:
## Summary
Add a node runner to help with nodes like Upload to kv and Join metadata upload. Also yanked out some metadata related code that is seemingly unused [slack thread](https://zipline-2kh4520.slack.com/archives/C072LUA50KA/p1752084499556549)

Tested by triggering the main method of the runner and confirmed that I see the export jobs being submitted. The plan is for the orch to directly call the run method as it does need to create the runner with the right Api class. 

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Introduced a utility method to check if a character sequence is empty.
  * Added a new runner for uploading data to a key-value store or join metadata, configurable via command-line arguments.
  * Added related command-line argument constants and environment variable for upload clients.

* **Bug Fixes**
  * None.

* **Refactor**
  * Removed support for the "NameByTeam" metadata endpoint and related logic from online and Spark components.

* **Tests**
  * Added comprehensive tests for the new key-value upload runner, covering normal operation and error handling.

* **Chores**
  * Updated existing tests to remove checks for the deprecated "NameByTeam" endpoint.


online/src/main/scala/ai/chronon/online/runner/KVUploadNodeRunner.scala
```diff
@@ -0,0 +1,182 @@
...
+
+      implicit val partitionSpec: PartitionSpec = PartitionSpec.daily
+      val range = Some(PartitionRange(null, endDs))
+
+      val runner = new KVUploadNodeRunner(api)
+      runner.run(metadata, nodeContent, range)
+    }
+  }
+
+  private def loadNodeContent(confPath: String, confType: String): NodeContent = {
```
tchow-zlai, line 152: The file locatd at confPath should actually be a `NodeContent` serialization I believe. We should be able to just `ThriftJsonCodec.fromJsonFile[NodeContent](confPath, check = false)` directly without having to specify a `confType` string

spark/src/main/scala/ai/chronon/spark/kv_store/KVUploadNodeRunner.scala
```diff
@@ -0,0 +1,154 @@
...
+
+import scala.util.{Failure, Success, Try}
+
+class KVUploadNodeRunner(api: Api) extends NodeRunner {
+
+  @transient lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  def createFetchContext(): FetchContext = FetchContext(api.genKvStore, MetadataDataset)
+
+  override def run(metadata: MetaData, conf: NodeContent, range: Option[PartitionRange]): Unit = {
```
tchow-zlai, line 21: feels like it's not necessary to have `metadata` as a method param in the abstraction as we should be able to just grab it from the conf later, and we can do case matching just once.  I'll make an adjustment to BatchNodeRunner.

tchow-zlai (approved):

spark/src/main/scala/ai/chronon/spark/kv_store/KVUploadNodeRunner.scala
piyush-zlai: yeah it is included in the nodecontent - have to reach a couple of levels down. Makes sense to yank though

PR #976, opened 2025-07-16
Bugfix edge case in union join with GB derivations

varant-zlai:
## Summary

Fixing bug where GroupBy derivations can erase left columns in the single join part union join case.

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured that original columns from the left table are preserved in the final join output when applying GroupBy derivations.

* **Tests**
  * Added a test to verify that the final join output schema includes both original and derived columns when using GroupBy derivations.


tchow-zlai (approved):

spark/src/main/scala/ai/chronon/spark/join/UnionJoin.scala
```diff
@@ -180,7 +180,14 @@ object UnionJoin {
     // Apply GroupBy derivations to the aggregated results if we're producing final join output
     // Else, it gets applied later in the JoinPartJob
     if (groupBy.hasDerivations && produceFinalJoinOutput) {
-      val finalOutputColumns = groupBy.derivationsScala.finalOutputColumn(baseResultDf.columns)
+      // In the case that we're producing final join output correctly, we don't want the derivations to erase base
+      // left DF columns. So we leverage the `ensureKeys` functionality on finalOutputCols to ensure that they are preserved
+      val ensureCols = if (produceFinalJoinOutput) {
```
tchow-zlai, line 185: we don't need this if condition right? By the time we get here `produceFinalJoinOutput` is always true.

varant-zlai: Oh you're right

nikhil-zlai (approved):
neat fix!

PR #1048, opened 2025-08-09
feat: Scheduling api ZIP-902

tchow-zlai:
## Summary

- Scheduling thrift API for end user interface.
- `ScheduleDeployRequest` and `ScheduleDeployResponse` are upserts. It will upsert a schedule AND turn it on. 
- Schedules are keyed on `branch` + `confName`. If the hash changes, they have to version up (thus changing the confName) before scheduling. 

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Introduced new scheduling operations for deploy, cancel, delete, and list actions within the orchestration service.
  * Enhanced scheduling capabilities with support for branch, configuration names, schedule identifiers, and pagination options.


api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,55 @@ struct NodeStepRunInfo {
...
+    2: optional string confName
+}
+
+struct ScheduleDeleteResponse {
+    1: optional string branch
+    2: optional string confName
+}
+
+struct ScheduleListRequest {
+    1: optional i64 limit
```
sean-zlai, line 150: nit: Use `i32` like `WorkflowListRequest`.

If we have a lot of these, would be nice to also have `offset` to support pagination, but the response would also need to return `total`, then I could [display](https://svelte-ux.techniq.dev/docs/components/Pagination) it.  Making sure if limit isn't passed it always returns all for MVP would be more than sufficient (can also pass in 1,000 or something)

api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,55 @@ struct NodeStepRunInfo {
...
+}
+
+struct ScheduleListDescription {
+    1: optional string workflowId
+    2: optional string confName
+    3: optional string branch
+    4: optional string hash
+}
+
```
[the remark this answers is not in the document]
sean-zlai: any other useful metadata to add for context, such as when the schedule was created/add, last ran, success of last run, etc?

I think this is mostly on the side of workflows being triggered, but might be nice to have an associations such as a workflow knowing the scheduleId that triggered it).

api/thrift/orchestration.thrift
tchow-zlai: I'll add the values. we can leave them alone for now

api/thrift/orchestration.thrift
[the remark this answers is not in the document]
tchow-zlai: Updated with a couple more fields, will add as we go.

api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,64 @@ struct NodeStepRunInfo {
...
+struct ScheduleCancelRequest {
+    1: optional string scheduleId
+}
+
+struct ScheduleCancelResponse {
+    1: optional string scheduleId
+    2: optional string branch
+    3: optional string confName
```
tchow-zlai, line 139: mode

api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,64 @@ struct NodeStepRunInfo {
...
+}
+
+struct ScheduleDeleteRequest {
+    1: optional string scheduleId
+}
+
+struct ScheduleDeleteResponse {
+    1: optional string scheduleId
+    2: optional string branch
+    3: optional string confName
```
tchow-zlai, line 149: mode

api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,64 @@ struct NodeStepRunInfo {
...
+}
+
+struct ScheduleListRequest {
+    1: optional i32 limit
+    2: optional i32 offset
```
tchow-zlai, line 154: branch

api/thrift/orchestration.thrift
```diff
@@ -115,6 +115,64 @@ struct NodeStepRunInfo {
     9: optional JobTrackingInfo jobTrackingInfo
 }
 
+// ====================== Start Orchestration Scheduling ======================
+struct ScheduleDeployRequest {
+    1: optional string branch
+    2: optional string mode
+    3: optional Conf conf
```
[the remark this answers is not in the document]
nikhil-zlai: oh the hash could change if the user updates - hmm. conf name should do then actually?

nikhil-zlai (approved):

PR #1071, opened 2025-08-13
feat: Add mode schedules map

tchow-zlai:
## Summary

- To correctly display online and offline schedules as well as indicate _what_ schedules to deploy, we need to pass in a map of conf modes to schedules defined by a conf in both the deploy submission request as well as when we retrieve conf details.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added per-mode scheduling to deployment and configuration APIs, enabling granular schedules by execution mode.
  * New optional scheduling options appear in deployment requests, configuration list items, and configuration detail responses.
  * Backward compatible: existing workflows remain unchanged unless the new options are used.
  * Improves flexibility for rollout planning and mode-specific timing.


piyush-zlai (approved):

api/thrift/orchestration.thrift
```diff
@@ -119,6 +120,7 @@ struct ScheduleDeployRequest {
     1: optional string branch
     2: optional string confName
     3: optional string confHash
+    4: optional map<planner.Mode, string> modeSchedules
```
sean-zlai, line 123: By using `planner.Mode` I assume we'll continue to embrace `Backfill` and `Deploy` vs `Offline` and `Online` naming (I think both will always live on in various aspects).

I can do the mapping on my side, but there are cases where we pass the opposite (and know we had a CLI verb naming discussion a few days ago).

sean-zlai: I like discrete naming with an enum vs wildcard strings any day :)

PR #1210, opened 2025-10-07
fix: check if eligible standalone and run union join

nikhil-zlai:
## Summary
currently we would hit a dead end if the join in question isn't eligible for union join

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- New Features
  - Introduces an automatic standalone union-join path that can skip final output when eligible.

- Performance
  - Early range validation to skip empty work.
  - Fewer intermediate writes and faster backfills/temporal joins for skew-free eligible runs.

- Refactor
  - Skew-free behavior exposed via a new config flag (enabled by default) with fallback to existing flow.

- Observability
  - Improved logging, progress updates, and latency/partition metrics.


spark/src/main/scala/ai/chronon/spark/join/UnionJoin.scala
```diff
@@ -218,8 +214,15 @@ object UnionJoin {
     }
   }
 
+  def isEligibleForStandaloneRun(joinConf: api.Join): Boolean = {
+    joinConf.left.isSetEvents &&
+    joinConf.getJoinParts.size() == 1 &&
+    !joinConf.isSetBootstrapParts &&
+    !joinConf.isSetLabelParts
```
tchow-zlai, line 221: should we just ignore labelParts since they aren't a thing anymore?

nikhil-zlai: i was thinking about it, thought that cleanup would happen elsewhere also. so didn't mess with that assumption here to be consistent with rest of the code base.

tchow-zlai (approved):

varant-zlai (approved):

PR #1232, opened 2025-10-15
Enhance TopicInfo parsing, add metrics logging, and refactor consistency job

cceasy:
## Summary
This MR introduces three minor changes:
1. Enhanced TopicInfo parsing to handle slash in parameter values
some config value may contains slash such as `ssl.truststore.location`, we need to support escape slashes on the uri, here we use double slash to escape slash in config value. Example: kafka://topic/host=X/ssl.truststore.location=//etc//kafka//truststore.jks
2. Added metrics logging with configurable prefix support
Added global metric prefix support via METRIC_PREFIX environment variable
3. Refactored ConsistencyJob to separate comparison table building logic
make buildComparisonTable as a standalone run mode (we nee this so that we can run our customized comparison which is much more faster than the consistency-metric-compute job)

All changes maintain backward compatibility, test cases added to cover the change.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added a "build-comparison-table" CLI subcommand/run mode and corresponding run-mode mapping.
  * Introduced a global metrics prefix affecting metric naming.

* **Bug Fixes**
  * Corrected topic parameter parsing to handle escaped double-slashes and empty values.
  * Consistency sampling: defaults to 100% when unset and skips processing when set to 0.

* **Chores**
  * Enhanced logging around topic parsing and metrics reporter setup.

* **Tests**
  * Added a test covering double-slash topic parsing.

### DIFF
--- a/online/src/main/scala/ai/chronon/online/DataStreamBuilder.scala
+++ b/online/src/main/scala/ai/chronon/online/DataStreamBuilder.scala
@@ -24,6 +24,7 @@ import org.apache.spark.sql.DataFrame
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+import java.security.MessageDigest
 import scala.collection.Seq
 import scala.util.Failure
 import scala.util.Success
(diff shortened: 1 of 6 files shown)


nikhil-zlai (changes requested):
can you split the PR - one for TopicInfo another for consistency job.

can you also add a unit test for your topic info case. 

can you also share the approach of your other consistency job?

PR #1245, opened 2025-10-23
Add planner support for chaining

piyush-zlai:
## Summary
Adding support in our planner for chaining of Joins to GBs. For the deploy scenario, we need to handle adding a dependency on the upstream parent join's metadata upload - this is needed by the downstream streaming GB job (as it looks up the join as part of its execution logic). If the downstream GB is not streaming, this dep is taken at the join metadata level to ensure the final join metadata upload is done after the parent one.

Example:
Src -> GB1 -> Parent Join -> Streaming GB -> Join
(in this case the streaming GB needs the parent metadata upload prior to running to call fetch join in the job)

Src -> GB1 -> Parent Join -> Snapshot GB -> Join
(in this case the final join needs the parent metadata upload completed)

To test this out along with some chaining Flink changes (coming in a follow up PR once I work through issues), I've added a simple canary chained GB and Join example that I used - 
```
$ zipline hub run-adhoc --conf compiled/joins/gcp/demo_chaining.downstream_join__0 --end-ds 2025-10-20
```

## Checklist
- [X] Added Unit Tests
- [ ] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Enabled GroupBy chaining via JoinSource so downstream GroupBys can use upstream join outputs
  * Improved streaming dependency handling for chained GroupBy/join pipelines

* **Tests**
  * Added extensive unit tests covering GroupBy/Join chaining, streaming vs non‑streaming behaviors, and dependency semantics
  * Added canary fixtures and compiled manifests for chained GroupBy and Join scenarios in GCP environments


api/src/main/scala/ai/chronon/api/planner/MonolithJoinPlanner.scala
```diff
@@ -44,33 +44,57 @@ case class MonolithJoinPlanner(join: Join)(implicit outputPartitionSpec: Partiti
...
+        Seq.empty
+      } else {
+        Option(groupBy.sources)
+          .map(_.asScala)
+          .getOrElse(Seq.empty)
+          .filter(_.isSetJoinSource)
+          .map { source =>
+            val upstreamJoin = source.getJoinSource.getJoin
+            val upstreamMetadataUploadTable = upstreamJoin.metaData.outputTable + "__metadata_upload"
+            new TableDependency()
```
tchow-zlai: is it worth moving any of this into [`TableDependencies`](https://github.com/zipline-ai/chronon/blob/main/api/src/main/scala/ai/chronon/api/planner/TableDependencies.scala)?

api/src/main/scala/ai/chronon/api/planner/TableDependencies.scala
```diff
@@ -126,4 +126,22 @@ object TableDependencies {
 
   }
 
+  def fromJoinSources(sources: java.util.List[api.Source]): Seq[TableDependency] = {
```
tchow-zlai: I was thinking this would be folded into `def fromSource` given that method has some handling around offset + lag that we may want to incorporate into the joinSource?

piyush-zlai: Chatted offline - will tackle this in a follow up pass. Hit some source related issues on some of the other batch nodes in the graph that I'll need to work through. We just need the metadata node for the deploy side so can unblock this right now and get to the other stuff later.

nikhil-zlai (approved):

PR #1267, opened 2025-10-31
Extend thrift + python APIs to support invoking model inference

piyush-zlai:
## Summary
Add support to express models + transforms in the thrift & python apis:
* Update our model construct to match what we sketched out in our [Model Support](https://docs.google.com/document/d/1m9pwQuNw2SpCuGQl9LNtXNTHQRV2Bt1mo30fCHq1NEI/edit?tab=t.0) doc.
* Add a `ModelTransforms` node - this will be analogous to GroupBys and Joins and it can be used express taking a Source, and applying n Models (included an example of how these would look for real in our canary py folder)
* Skipped publishing models to Hub in this first cut - Hub currently doesn't support model objects being sent and we throw an exception if they're in the list, so filtering them for now and we can add support in a follow up. 

## Checklist
- [ ] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added Vertex AI and SageMaker backends, inference specs, resource configs, and ModelTransforms to compose models with sources and passthrough fields.

* **Refactor**
  * Reworked model configuration to inference-spec-based models and ModelTransforms.
  * Centralized source normalization used across group-by and model transforms.
  * JoinSource construction now accepts an optional query parameter.

* **Chores**
  * Excluded model configs from local hash/diff computations to streamline upload summaries.


varant-zlai (approved):
Small nits but otherwise LGTM

PR #1271, opened 2025-11-03
feat: monolith-join fixes for offline/online parity

nikhil-zlai:
## Summary

This PR fixes several bugs in the modular monolith join pipeline that caused offline/online feature value mismatches, and adds test infrastructure to catch regressions.

### Bug fixes

**`JoinPartJob`: fix left column selection with key mappings** (`rightToLeft.values` was `.keys`)
When a join part has a key mapping (e.g. left key `user_id_new` → right key `user_id`), `JoinPartJob` was selecting the wrong columns from the left table — right key names instead of left key names. This caused `AnalysisException` for any join with a non-trivial key mapping.

**`ModularMonolith`: null-start guard in `computeNodeRange`**
No-agg GroupBys (e.g. point-in-time lookups with no window) return a null `start` in their `DateRange`. Without a null guard, computing the union of input ranges threw a `NullPointerException`. Now falls back to `queryRange.start` when all starts are null.

**`ModularMonolith`: keep `alignOutput=false` in `JoinPartJob`**
SNAPSHOT join parts write to partition D-1. `MergeJob` reads at D-1. Setting `alignOutput=true` caused join parts to write to D, which MergeJob couldn't find, producing offline/online diffs.

**`FetcherTestUtil`: read derivation output table for offline comparison**
`JoinPlanner` writes the `JoinDerivationJob` output to `outputTable + "__derived"`, not `outputTable` (which holds the pre-derivation MergeJob output). `compareTemporalFetch` was reading from the wrong table, causing ~4500 spurious diffs per fetcher test run for joins with derivations.

**`SparkSessionBuilder`: set `spark.driver.host=127.0.0.1` in local mode**
With only `spark.driver.bindAddress=127.0.0.1` set, the block manager registered its address using the machine's hostname (e.g. `192.168.0.104`) while the socket was bound to localhost. This caused `TaskResultLost` failures in local-mode tests when tasks tried to fetch results from the block manager.

### Test infrastructure

**`FetcherDataFrameGenTest`**: new fixed-date DataFrameGen test that reproduces the offline/online parity scenario with a stable, deterministic dataset. Uses timestamp shifting to pin all generated data to a fixed reference date (`2021-04-10`), making failures reproducible and debuggable across runs.

## Checklist
- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


nikhil-zlai (commented):
comments

varant-zlai (approved):

PR #1282, opened 2025-11-11
Support avro.java.string in Avro Schema

JoeTime17:
## Summary
Fix ClassCastException in Flink Avro Deserialization with avro.java.string Configuration

- Problem
Flink jobs processing Avro messages from Kafka with Schema Registry were failing with a ClassCastException when the Avro schema used "avro.java.string": "String" configuration:
`java.lang.ClassCastException: class java.lang.String cannot be cast to class org.apache.avro.util.Utf8`
This error occurred in SourceProjectionDeserializationSchema during the deserialization process when AvroConversions.toChrononRowCached() expected Utf8 objects but received String objects.

-   Root Cause
The issue was in AvroConversions.scala where the string converter function only handled Utf8 types. When Avro schemas include "avro.java.string": "String", the Avro library produces java.lang.String
objects instead of the default org.apache.avro.util.Utf8 objects, causing the type mismatch.

-   Solution
1.Changed type parameter in toChrononRowCached from Utf8 to Any to enable pattern matching
2.Added type-safe string handling with pattern matching to support both String and Utf8 types

-   Impact
1.Backward compatible - existing Utf8 handling continues to work
2.Forward compatible - now supports modern avro.java.string configurations
3.Fail-fast behavior - provides clear error messages for unexpected types
4.Spark unaffected - Spark uses different abstraction layers that don't encounter this issue

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Avro deserialization to accept multiple string representations and provide clearer errors for unsupported types.

* **Tests**
  * Added tests verifying reliable string conversion across different Avro schema configurations, including mixed string representations.

### DIFF
--- a/online/src/main/scala/ai/chronon/online/serde/AvroConversions.scala
+++ b/online/src/main/scala/ai/chronon/online/serde/AvroConversions.scala
@@ -257,10 +257,11 @@ object AvroConversions {
       },
       {
         case avString: Utf8 => avString.toString
-        case str: String => str
-        case other => throw new IllegalArgumentException(
-          s"Unexpected string type: ${other.getClass.getName}. Expected String or Utf8, got: $other"
-        )
+        case str: String    => str
+        case other =>
+          throw new IllegalArgumentException(
+            s"Unexpected string type: ${other.getClass.getName}. Expected String or Utf8, got: $other"
+          )
       }
     )
   }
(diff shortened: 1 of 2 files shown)


piyush-zlai (approved):
Thanks for the fix + test!

PR #1289, opened 2025-11-12
fix: line up join inner table names

nikhil-zlai:
## Summary

we want all the names to be suffixed with __(version|hash)?__mode.

we also don't want backfill in the name of the node - a node can be either backfilled or frontfilled based on the name of the node

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added execution support for a union-join path and union-join nodes.

* **Enhancements**
  * Utilities to strip trailing version and optional team segments from names.
  * Standardized metadata layer labels and output name patterns (renamed backfill tags to group_by/monolith_join/staging; unified suffixes like __source, __merged, __derived, __staging).
  * Batch runner logs and executes union-join jobs.

* **Tests**
  * Relaxed and updated tests to match the new metadata naming conventions.


api/src/main/scala/ai/chronon/api/Extensions.scala
```diff
@@ -219,6 +220,26 @@ object Extensions {
         case None => None
       }
     }
+
+    def nameWithoutVersion: String = MetadataOps.splitTrailingVersion(metaData.getName)._1
+
+    def nameWithoutTeamOrVersion: String = {
```
tchow-zlai: am actually wondering if it's worth creating an internal `Name` case class something like 

```scala
case class Name(baseName: String, version: String, team: String) 
```

with a constructor `apply(nameStr: String)`

tchow-zlai (approved):

PR #1297, opened 2025-11-18
feat: batch-node-runner: archive tables on semantic changes

nikhil-zlai:
## Summary

before each node runs we check whether semantic hash matches, if not we archive the table. 

we also assign the semantic hash after the core run method of the node.

## Checklist
- [ ] Added Unit Tests TODO
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Batch runs now validate semantic hashes for output tables before and after execution; mismatches are archived/reused and hashes persisted.
  * Added utilities for table renaming, archival/reuse workflows, and semantic-hash management with timestamped archives.

* **Bug Fixes / Improvements**
  * Archival/drop operations now return archive names and handle failures more robustly.
  * SQL property keys/values are safely quoted/escaped.
  * Archive suffix centralized for consistent naming.

* **Tests**
  * New tests cover hash validation, archival flows, property preservation and edge cases.


cristian-zlai:
I think we should double down on iceberg.

From what I understand from https://iceberg.apache.org/docs/latest/branching/ and https://iceberg.apache.org/docs/latest/spark-ddl/#dropping-a-view
We can consider each semantic hash a branch. So we'd do operations like:
```
# Start the branch
ALTER TABLE db.table CREATE BRANCH IF NOT EXISTS test_branch;
# Clear the branch and populate with whatever you may want to bootstrap it with (or empty table i.e. drop all data)
INSERT INTO db.table.<semantic_hash>  ... 
```
After the branch is finished computing if this is going on on the main branch then we update the table:
```
CALL catalog_name.system.fast_forward('prod.db.table', 'main', '<semantic_hash>');
```
Other alternatives here are create an empty version of the table (initially), and tag it as empty. Then when we create the branch / new semantic hash, we can do 
```
CREATE BRANCH IF NOT EXISTS <semantic_hash> as of tag 'empty'
```

Some advantages here:
* Cheap -> Snapshots can be regulated in terms of retention and any other governance policies.
* Clean -> The data warehouse looks good, since there's only 1 table in theory.
* Decoupled -> No DDL restrictions between snapshots. 
* Rollbacks -> If something broke during an update and you roll back two or whichever many semantic hashes past, it picks up from the last snapshot. 
* User friendly -> The user only needs to know about the main table. All the semantic hash and pointer management is handled by the engine.

nikhil-zlai:
@cristian-zlai and I discussed offline - and it seems like the branches don't support their own table properties. Will go with table archiving approach for now.

cristian-zlai (approved):

PR #1299, opened 2025-11-18
feat: make the derived join output tables easy to consume downstream

nikhil-zlai:
## Summary

adding a derivation to a join without derivations currently produces a new table with a _derived prefix.

however that makes it hard to consistently consume it downstream. we instead un_prefix the terminal table

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Refactor**
  * Merge outputs now receive a "__merged" suffix when derivations exist.
  * Derivation artifacts are only created when derivations are present, reducing clutter.
  * Derivation outputs no longer use a "__derived" suffix; naming and metadata consistently reference the final merge output.

* **Chores**
  * Metadata/execution info is now recorded and propagated with merge/derivation steps.

* **Tests**
  * Updated tests to validate the new naming and execution metadata flows.

### DIFF
--- a/api/src/main/scala/ai/chronon/api/planner/JoinPlanner.scala
+++ b/api/src/main/scala/ai/chronon/api/planner/JoinPlanner.scala
@@ -150,7 +150,7 @@ class JoinPlanner(join: Join)(implicit outputPartitionSpec: PartitionSpec)
         "merge",
         mergeNodeName,
         deps,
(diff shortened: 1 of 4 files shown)


nikhil-zlai:
closing this since we decided to keep the names as is - merge job produces table: <join_name>, derivations jobs produces table: <join_name>_<derived>

in python we need to have a helper for join.derived_table just like join.table.

PR #1300, opened 2025-11-18
UNIQUE_TOP_K last seen mode

abbywh:
## Summary

Add an option to sort either on `FIRST_SEEN` or `LAST_SEEN` for any given unique_id

## Checklist
- [ x ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Release Notes

* **New Features**
  * Added collision strategy parameter to UNIQUE_TOP_K aggregations with two modes: IGNORE (default) and UPDATE. UPDATE mode replaces duplicates with the same ID with the latest value, enabling last-seen deduplication semantics across Int, Long, String, and Struct data types.

* **Documentation**
  * Updated UNIQUE_TOP_K documentation to explain collision strategy options and usage examples.

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/base/UniqueOrderByLimit.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/base/UniqueOrderByLimit.scala
@@ -120,29 +120,29 @@ object UniqueOrderByLimit {
     }
   }
 
-  // Lenient (LAST_SEEN) implementation that allows replacing elements with same ID
-  def initLenientState[T, OrderType]: LenientState[T, OrderType] =
-    LenientState(new util.ArrayList[T](),
-                 new util.HashMap[Long, Int](),
-                 null.asInstanceOf[OrderType],
-                 indicesDirty = false)
+  // UPDATE collision strategy implementation that allows replacing elements with same ID
+  def initUpdateState[T, OrderType]: UpdateState[T, OrderType] =
+    UpdateState(new util.ArrayList[T](),
+                new util.HashMap[Long, Int](),
+                null.asInstanceOf[OrderType],
+                indicesDirty = false)
 
-  case class LenientState[T, OrderType](elems: java.util.ArrayList[T],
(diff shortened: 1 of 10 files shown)


thrift/api.thrift
```diff
@@ -217,6 +217,16 @@ enum Operation {
     UNIQUE_TOP_K = 20  // returns top k unique elements based on frequency
 }
 
+/**
+ * Deduplication mode for UNIQUE_TOP_K operation.
+ * Determines behavior when duplicate unique_id values are encountered.
+ */
+enum DedupMode {
+    // Keep the first occurrence when a duplicate unique_id is seen
+    FIRST_SEEN = 0
```
nikhil-zlai: The naming a bit overloaded with first and last aggregations which indicate a time based sorting. 

```
"collision_strategy" - "update, ignore"
"uniqueness" - "strict, lenient"
```

aggregator/src/main/scala/ai/chronon/aggregator/base/UniqueOrderByLimit.scala
```diff
@@ -120,4 +120,166 @@ object UniqueOrderByLimit {
...
+    }
+
+    def insert(elem: T, state: LenientState[T, OrderType]): Unit = {
+      val elemId = getId(elem)
+      val orderKey = getOrderKey(elem)
+
+      // LAST_SEEN behavior: if ID exists, replace the element at that index
+      if (state.idToIndex.containsKey(elemId)) {
+        val existingIndex = state.idToIndex.get(elemId)
+        state.elems.set(existingIndex, elem)
```
nikhil-zlai: this will be non-deterministic I think - if the inputs are not time sorted. Which is the case in the non-union-join node and in batch accurate features.

thrift/api.thrift
abbywh: I think 

`collision_strategy` is good, I could see us eventually add "best" and "worst"

PR #1311, opened 2025-11-25
Model training & deploy - thrift changes + py examples

piyush-zlai:
## Summary
Rough sketch of what our thrift api would look like to support model training & deploys based on [Model support](https://docs.google.com/document/d/1m9pwQuNw2SpCuGQl9LNtXNTHQRV2Bt1mo30fCHq1NEI/edit?tab=t.0#bookmark=id.fp8lgkzb1jup). Took a stab at the Python api updates and a sample example ctr model config to make things a bit more concrete. As we integrate this in subsequent PRs we can refine and iterate on the thrift & py apis. 

## Checklist
- [ ] Added Unit Tests
- [X] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Extended Model API with training & deployment specs, rollout strategies, and model artifact URI.
  * Added a full click‑through‑rate reference model, runnable trainer/example package, and updated demo joins/derivations for CTR enrichment.
  * Added a user-facing README for the CTR example.

* **Bug Fixes / Improvements**
  * Centralized window normalization for consistent handling of time-window inputs.

* **Chores**
  * Removed an outdated derivation and bumped derivation versions.


python/src/ai/chronon/model.py
```diff
@@ -14,6 +16,12 @@ class ModelBackend:
     SAGEMAKER = ttypes.ModelBackend.SageMaker
 
 
+class DeploymentStrategyType:
+    BLUE_GREEN = ttypes.DeploymentStrategyType.BLUE_GREEN
```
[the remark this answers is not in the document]
tchow-zlai: nvm see it in the thrift.

thrift/api.thrift
```diff
@@ -590,3 +598,81 @@ struct ModelTransforms {
...
+    // deploys the model immediately to the endpoint without any traffic ramping
+    IMMEDIATE = 2
+}
+
+struct Metric {
+    1: optional string name
+    2: optional double threshold
+}
+
+struct RolloutStrategy {
```
tchow-zlai: ooc which portion of this arch will conduct the actual rollout? Or is that something that we haven't settled on?

thrift/api.thrift
```diff
@@ -590,3 +598,81 @@ struct ModelTransforms {
     // If the source is a JoinSource, this can be skipped as we look it up from the join metadata
     5: optional TDataType keySchema
 }
+
+struct TrainingSpec {
```
nikhil-zlai: does this need a trainer python class / module name?

nikhil-zlai (approved):
added comments, but approving to establish that my comments are non-blocking.

thrift/api.thrift
piyush-zlai: This will be done by the specific model platform under the hood. Vertex and Sagemaker provide slightly different rails for this. Vertex requires you to do a bit more on your own, Sagemaker seems to actually support b/g and other primitives for model deploys to endpoints.

python/test/canary/models/gcp/click_through_rate.py
```diff
@@ -0,0 +1,89 @@
...
+            max_replica_count=1,
+            machine_type="n1-standard-4"
+        ),
+        job_configs={
+            "epochs": "1000",
+            "batch-size": "32",
+            "learning-rate": "0.001",
+        }
+    ),
+    deployment_conf=DeploymentSpec(
```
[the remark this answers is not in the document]
piyush-zlai: it primarily covers how we want to deploy the model - which container image to use for inference, which Vertex / Sagemaker endpoint to deploy the model to, rollout strategy etc. There is some overlap with inference here ofc like the container cfg + number of resources for serving but I think the deployment name seems more pertinent..

python/test/canary/models/gcp/click_through_rate.py
```diff
@@ -0,0 +1,89 @@
...
+        "'listing_tags', listing_id_tags, 'listing_category', listing_id_primary_category)",
+    },
+    output_mapping={
+        "ctr": "score"
+    },
+    # captures the schema of the model output
+    value_fields=[
+        ("score", DataType.DOUBLE),
+    ],
+    training_conf=TrainingSpec(
```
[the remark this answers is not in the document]
piyush-zlai: yeah should have called this out a bit more. There's a couple of options here:
* Users define their actual training logic in files in their Chronon repo. In this case we do need to either build a Python Source distro (Vertex) or a flattened tarball (Sagemaker) or a docker image (Vertex / Sagemaker) for them. If we go down this path, I'd imagine a folder called 'models' in the repo and a sub-dir per model. In that subdir the user defines their py definitions and requirements etc and we bundle the sub-dir into an archive / docker img. 
* Users define this themselves outside and build the custom container img and list it here. 

I will refine this a bit further after experimenting with these a bit more. If we can do the Python tarball approach with the requirements file then we could have the users define the framework image (xgboost x or pytorch y) they want to use, the tarball coordinates (or we build) etc.

thrift/api.thrift
piyush-zlai: Based on my read of the docs, if we're going with a custom docker image the training script is typically the entrypoint of your image (e.g. - https://docs.cloud.google.com/vertex-ai/docs/training/create-custom-container). If we do pursue the python src dist / tarball way with prebuilt containers, then I think we do need the entrypoint script name so we'll need it there.

david-zlai (approved):
looks fine. don't see any backwards incompatible changes to existing abstractions

PR #1312, opened 2025-11-26
Fix BigTable list and multiGet filter overwrite bug

cceasy:
## Problem
When we use chronon fetcher service to fetch join features, we got unexpected result when do bulk fetch, some join features missing streaming results.

The issue was in BigTableKVStoreImpl.readRowsMultiGet() and BigTableKVStoreImpl.list(), the query filters would be overwritten.

When fetching multiple entities with different time ranges in a single multiGet call, the BigTable Query object's timestamp filter was being overwritten, causing incorrect results. This occurred because the BigTable Query API only allows setting one filter at a time - setting a new filter replaces the previous one.

## Solution
- Group requests by time range before creating queries to avoid filter conflicts
- Use chained filters via Filters.FILTERS.chain() to properly combine multiple filter conditions
- Apply the same fix to the list function 
- Create separate queries for each time range group, ensuring each group gets the correct filters

## Changes
1. BigTableKVStoreImpl.scala:
  - Modified readRowsMultiGet() to group requests by (startTsMillis, endTsMillis) for time-series tables
  - Implemented filter chaining for proper filter combination
  - Created helper method generateTimeSeriesRowKeys() for cleaner code
  - Applied similar fix to list() function
2. BigTableKVStoreTest.scala:
  - Added critical test case: "handle multiple entities with different time ranges in single query" - tests the exact scenario that would fail before the fix
  - Added edge case tests for mixed request types and missing data scenarios

## Summary by CodeRabbit

* **Performance Improvements**
  * Batched and grouped multi-get requests by dataset and time range for fewer BigTable calls and lower latency.
  * Constructed per-group queries and row-key sets to reduce redundant work.

* **Refactor**
  * Introduced time-series–aware filtering and unified query construction while preserving public APIs and metrics.

* **Tests**
  * Expanded coverage for multi-entity queries, mixed time-series/non-time-series scenarios, tiled datasets, and edge cases.


cceasy:
here the CI task failed, BUT the tests actually passed in my local
image

online/src/main/scala/ai/chronon/online/fetcher/GroupByFetcher.scala
```diff
@@ -172,7 +172,8 @@ class GroupByFetcher(fetchContext: FetchContext, metadataStore: MetadataStore)
     val allRequestsToFetch: Seq[GetRequest] = groupByRequestToKvRequest.flatMap {
       case (_, Success(LambdaKvRequest(_, _, batchRequest, streamingRequestOpt, _, _))) =>
         // If a batch request is cached, don't include it in the list of requests to fetch because the batch IRs already cached
-        if (cachedRequests.contains(batchRequest)) streamingRequestOpt else Some(batchRequest) ++ streamingRequestOpt
+        if (cachedRequests.contains(batchRequest)) streamingRequestOpt.toSeq
```
piyush-zlai: shouldn't need this right?

piyush-zlai (commented):
Dropped a suggestion on refactoring, let me know what you think.

online/src/main/scala/ai/chronon/online/fetcher/GroupByFetcher.scala
cceasy: yes, reset back

cloud_gcp/src/main/scala/ai/chronon/integrations/cloud_gcp/BigTableKVStoreImpl.scala
```diff
@@ -119,118 +119,148 @@ class BigTableKVStoreImpl(dataClient: BigtableDataClient,
...
   private def readRowsMultiGet(
       requestsByDataset: Map[String, Seq[KVStore.GetRequest]]): Seq[Future[Seq[KVStore.GetResponse]]] = {
-    requestsByDataset.map { case (dataset, datasetRequests) =>
+    requestsByDataset.flatMap { case (dataset, datasetRequests) =>
       val targetId = mapDatasetToTable(dataset)
       val datasetMetricsContext = tableToContext.getOrElseUpdate(
         targetId.toString,
         metricsContext.copy(dataset = targetId.toString)
       )
+      val tableType = getTableType(dataset)
```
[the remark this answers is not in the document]
cceasy: good suggestion, function `readRowsMultiGet` has too many nestings, now it's clear after this refactor

piyush-zlai (approved):
Thanks for the fixes and updates to the tests!

PR #1318, opened 2025-12-03
feat: entity compiling and exposing via CLI

cristian-zlai:
## Summary

We add a concept of `Entity` which can be used for defining and semantically bonding different group bys and joins.

For this the approach is to be as backwards compatible as possible while providing the visibility required. 
If entities are not used for definition, it works as well as before.
If entities are used during definition, we register these usages and then after compiling can provide a json or plaintext of the features per entity.

TODO: See if there's a less prone to mistakes way of associating a source to an entity (right now it's being done during definition. 

TODO: csv output.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Release Notes

* **New Features**
  * Introduced entity registration framework to define and manage feature entities
  * Feature configurations now accept entity references for keys and field mappings
  * Added --entity-summary and --entity-csv CLI options to export entity metadata during compilation

### DIFF
--- a/python/src/ai/chronon/cli/compile/compile_context.py
+++ b/python/src/ai/chronon/cli/compile/compile_context.py
@@ -20,6 +20,7 @@ from ai.chronon.cli.compile.display.compiled_obj import CompiledObj
 from ai.chronon.cli.compile.serializer import file2thrift
 from ai.chronon.cli.formatter import Format
 from ai.chronon.cli.logger import get_logger, require
+from ai.chronon.repo.entity_register import EntityRegister
 
 logger = get_logger()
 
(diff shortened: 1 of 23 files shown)


cristian-zlai:
@coderabbitai review

nikhil-zlai:
small modifications: 

- we should add a top level "entities.py" file where users define the `type` and `docstring` and name of the `Entity`. This is the only concept added to the SDK.

- In the api, if users don't use a select - we select the default name, other wise we automatically form an "EntityAssociation" list during compile time walking based on entity.select('expr'), no need for table here - we look at the tables along side the query in the source instead.

If we want this to be even more seamless, we don't even require the new import - just use string matching in groupBy key -> entities.py to make the associations.

cristian-zlai:
1. I put it in the entities folder, but could just as well be a top level file.
2. String matching can be tricky as some entities use the same strings ("id", "user_id" -> guest / host), while it could be a quality of life to use it when no ambiguity, it also feels like a bit of a gotcha. 
3. Will work on getting the table out of the select. 

One thing not exposed here is after the compiling there's an entity summary that is useful. Similar to the table used to power the feature view. I'm open to ideas on what do we do with that info, do we build a thrift object? do we pass it into the metadata of the object built (gb.metaData.entities: Map[string, string]), etc.

PR #1324, opened 2025-12-06
feat: stats compute + summarize cardinality logic

cristian-zlai:
# Summary

Enhance the StatsCompute logic to take into account cardinality.

## Main changes (core) :
* We subclassed StatsCompute to EnhancedStats which has the categorical treatment. Eventually we should remove the OG version there's a lot of repetition
* Now we add metadata to the KvRDD that contains which features are categorical (cardinalityMap) as well as the required fields in order to reconstruct the aggregator (inputSchema-ish)
* Added an EnhancedStatsStore as an interface with upload (and originally fetching) but later moved the fetching logic to online where it makes more sense. This could be revised.
* One big refactor here is that instead of using the rowAggregator we should just compute via sql and build the rowAggregator IRs based on the sql output. There's no need to go row by row computing these IRs which are fairly simple. This is the main benefit of Summarizer over StatsCompute. 

## Orchestration:
* Added a node after backfill that does the stats compute. This is under a feature flag. This means I also had to change the python API. 
* There's some logic in there that could be better placed elsewhere since BatchNodeRunner is already pretty overloaded. We can revise post "demo" since this should be a simple refactor.

## Online:
Having uploaded the data into the Kv store we need to fetch it "merged". To fetch merged we need to get the IRs build aggregators, fetch tiles and merge. This requires rebuilding quite a few schemas and operations. Here definitely we need some TTLCache magic.  We also add derived metrics. 

## Service:
Added endpoints that leverage the JavaStatsService from the online module. The service logic is pretty calm. Most of the good stuff is in online module.

## Known issues:
* Warm up the KV store before fetching else it says the schema may not been uploaded but it works on the 3rd / 4th time. 
* Step days is set to 1 but there's no reason why it couldn't be larger...
* There's no namespacing in the stats dataset. (So we interact with each other for now)

## Followups:
* We can't add histogram to booleans due to type... we should try enable those.
* FrequentItems normalization denormalization did not work when adding it to the StatsGenerator metrics we were computing. 
* We can namespace some of the stats by adding a prefix.
* The Impl for the KVStore is not complete.
* TTLCache to avoid hitting the KVStore.
* Impl parameters could be better, we don't need multiple mutations for a cell since we add the stamp to the keys.
(according to [docs](https://docs.cloud.google.com/bigtable/docs/schema-design-time-series) )
* Handling of CardinalityMap needs to be improved (daily cardinality map that gets merged over time? -> only latest? -> One pass and stick with it? )

Testing Locally:
1. You need the Jar from this branch. So 
```
./scripts/distribution/build_and_upload_artifacts.sh --customer_ids dev --gcp --skip-checks --skip-wheel
```
2. You need to enableStatsCompute on the conf.  (Using the python changes in the branch -- ./mill python.installEditable
```python
v1 = Join(
    left=source,
    row_ids=["event_id"], # TODO -- kill this once the SPJ API change goes through
    right_parts=[
        # User behavioral features (aggregated over time windows)
        JoinPart(
            group_by=user_activities.v1,
        ),
        # Listing dimension attributes (point-in-time lookup)
        JoinPart(
            group_by=dim_listings.v1,
        ),
        # Listing dimension attributes (point-in-time lookup)
        JoinPart(
            group_by=dim_merchants.v1,
            prefix="merchant_"
        ),
    ],
    version=7,
    online=True,
    output_namespace="data",
    step_days=2,
    enable_stats_compute=True,
)
```
3. You also need to update teams.py to use your dev jar (from step 1)
4. Run local orchestrator and execute a backfill (the node should appear then)
5. After the stats node finished. You should be able to see data in KV Table. You can check that in the cloud provider if needed. Next we need to fetch. Fetching comes from fetcher service. So build that docker image (./scripts/distribution/publish_docker_images.sh --local), and then run it (docker run or gcp_compose... see docker/fetcher/README.md)
6. Run those fetches.
curl "http://localhost:9000/v1/stats/gcp.demo.v1__7?startTime=1764115200000&endTime=1765310957443" 

### Example

Example output:
```shell 
curl "http://localhost:9000/v1/stats/gcp.demo.v1__7?startTime=1764115200000&endTime=1765310957443" | jq | pbcopy
```

```json
{
  "success": true,
  "tableName": "gcp.demo.v1__7",
  "tilesCount": 1,
  "statistics": {
    "user_id_view_event_sum_30d_min": null,
    "user_id_favorite_event_sum_7d_approx_unique_count": null,
    "listing_id_price_cents_min": 1283,
    "user_id_add_to_cart_event_average_14d_approx_percentile": null,
    "user_id_view_event_average_7d_average": null,
    "user_id_favorite_event_sum_7d__zero_sum": null,
    ...
    "listing_id_inventory_count_max": 483,
    "listing_id_primary_category_histogram": {
      "jewelry": 54,
      "electronics": 126,
      "books": 163,
      "toys": 137,
      "beauty": 124,
      "sports": 130,
      "clothing": 104,
      "automotive": 108,
      "food": 133,
      "home_garden": 89
    },
    "user_id_click_event_sum_7d_min": null,
    "listing_id_is_expensive_max": 1,
    "user_id_is_mobile_sum_14d_average": null,
    "user_id_is_desktop_sum_7d_average": null,
    "user_id_click_event_sum_1d_variance": null,
    "user_id_is_mobile_sum_7d_max": null,
    "user_id_favorite_event_average_7d__null_sum": 1262,
    "user_id_purchase_event_sum_1d_max": null,
    "row_id_approx_unique_count": 519
  },
  "timeRange": {
    "startTime": 1764115200000,
    "endTime": 1765310957443
  }
}
```

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Enhanced statistics: cardinality-aware metrics, zero-detection, derived fields (std_dev, false_sum), and new metric transforms.
  * New HTTP endpoint and Java-friendly service to fetch pre-computed table/time-range stats.
  * Storage & upload: KV-store backends and Spark path producing daily tiled enhanced stats; optional opt-in flag to enable stats compute within joins.
* **Tests**
  * New test suite validating enhanced metrics, cardinality detection, and KV serialization.
* **Documentation**
  * Added comprehensive development guide for the enhanced statistics system.


david-zlai (approved):
merging for now

PR #1342, opened 2025-12-12
Conditionally use derivation node for model transform

david-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency resolution for join sources with derivations — backfills now target the derived output when present, preserving prior behavior otherwise.
* **Tests**
  * Added and clarified tests covering join scenarios with and without derivations to ensure correct plan construction and dependency wiring.


piyush-zlai (approved):
not to block, but lets add tests to ModelTransformsPlannerTest as a follow up

PR #1347, opened 2025-12-12
Add training data window to table deps for training data node

david-zlai:
## Summary

When model training sets a training data window, we need to orchestrate to the upstream dependencies.

For example below, a model training spec that sets a 10 day window should require 10 days from its label staging query shown below. (Currently the code just does 1 day)

image
_**TODO: check if it's okay here that this is actually doing 10 + 1 = 11 days**_

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Improvements**
  * Training configurations can now include an explicit training data window (optional), replacing previous default behavior—improves accuracy of model training dependency and planning.
  * Table dependency calculations and handling for training flows are more robust when training info is missing or partial.

* **Tests**
  * Updated tests to validate training-data dependencies and window offsets more precisely.


piyush-zlai (approved):
Suggestion around structuring / null safety but otherwise lgtm

PR #1356, opened 2025-12-23
relax Backfill join temporal window boundary

JoeTime17:
## Summary

This PR fixes a bug in temporal join backfills where events occurring at the exact same timestamp as a query were incorrectly excluded from the aggregation window, resulting in NULL values instead of the correct aggregated results.

The bug affected both execution modes:
  - skewFree mode (spark.chronon.join.backfill.mode.skewFree=true)
  - non-skewFree mode (spark.chronon.join.backfill.mode.skewFree=false)

---
**Problem Description**

Issue: Temporal Window Boundary Inconsistency

In the documentation example at https://zipline.ai/docs/authoring_features/Join#left-side-events-right-side-streaming-events, for a left-side row with timestamp 2023-12-20 12:01:01.923, the aggregation window is defined as:
  - Range: [2023-12-10 12:01:01.923, 2023-12-20 12:01:01.923)
  - Excluded: The right boundary 2023-12-20 12:01:01.923 itself is not included

Inconsistency Between Online and Offline Behavior

Online Fetch (Streaming)
  - When a GroupBy key appears for the first time, aggregations like COUNT return a non-null value (e.g., 1)
  - This is because queryTs is always > event_time, so the current event is included in its own aggregation

  Backfill Join (Batch)
  - When a GroupBy key appears for the first time, aggregations return NULL
  - This is because the current event at timestamp T is excluded from the window [T-window, T)
  - Consequently, backfilled values are systematically off by 1 compared to online fetch values

---
**Root Cause**
Two different bugs in two execution paths:
  1. SkewFree mode (cumulateAndFinalizeSortedIterator method):
    - Used strict inequality sortedInputs(inputIdx).ts < sortedEndTimes(queryIdx).ts
    - Events at ts == queryTs were excluded
  2. Non-skewFree mode (cumulate method):
    - Binary search logic would skip forward past matching timestamps
    - Events at ts == queryTs were added to the next query's window instead

---
**Use Cases Affected**
This fix is especially important for:
  - Self-joins where left and right sources are the same table
  - Temporal accuracy joins with events at identical timestamps
  - Fraud detection and other use cases sensitive to first-occurrence detection

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [x] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed temporal window boundary handling to include events whose timestamps equal window boundaries, ensuring aggregations count and sum those events.

* **Tests**
  * Added test coverage verifying inclusion of events at window boundaries (new cases added to aggregation tests).

* **Documentation**
  * Updated example to reflect the inclusive boundary behavior (adjusted example timestamp).

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/SawtoothAggregator.scala
@@ -185,7 +185,7 @@ class SawtoothAggregator(aggregations: Seq[Aggregation], inputSchema: Seq[(Strin
 
     while (queryIdx < sortedEndTimes.length) {
 
-      while (inputIdx < sortedInputs.length && sortedInputs(inputIdx).ts < sortedEndTimes(queryIdx).ts) {
+      while (inputIdx < sortedInputs.length && sortedInputs(inputIdx).ts <= sortedEndTimes(queryIdx).ts) {
         queryIr = windowedAggregator.update(queryIr, sortedInputs(inputIdx))
         inputIdx += 1
       }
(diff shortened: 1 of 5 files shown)


nikhil-zlai (approved):
pretty rigorous fix! thanks for working on this!

nikhil-zlai:
reverting due to backwards compatibility issues - we need to re-add it with a flag. the default behavior should be to exclude.

PR #1379, opened 2026-01-13
perf: avoid creating IRs in a combine stage

nikhil-zlai:
## Summary

Sometimes the size of the IR could be larger than raw data. In those cases creating all the IRs via a combine-reduce pattern, via aggregate_by_key will fill up memory of the mapper very quickly. This is compounded by the fact that a mapper is not sharded by key and could contain all the keys in its combined state.

This PR creates a flag to toggle the combine-reduce behavior with pure reduce behavior. The downside of pure-reduce would be cases where there are legitimate hot-keys. Users must filter out hot-keys (via where clauses) in those cases.

We use scala Either to avoid initializing the IR on the mapper side.

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option (enabled by default) to switch aggregation behavior during group-by uploads, allowing an alternate combined aggregation path for optimized processing.

* **Chores**
  * Consolidated imports and cleaned up formatting.
  * Registered an additional type for serialization to improve compatibility.
  * Clarified internal comments for maintainability.


cristian-zlai (approved):
Love the flag behavior.

PR #1403, opened 2026-01-28
feat: Snowflake partition listing

tchow-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added an Azure format provider that selects Iceberg or Snowflake for tables.
  * Enabled vault-based private-key authentication for Snowflake access.

* **Removed**
  * Deprecated the previous Spark Snowflake format provider.

* **Configuration**
  * Updated runtime and test configs to set the new format provider and replace SNOWFLAKE_STORAGE_INTEGRATION with SNOWFLAKE_VAULT_URI.


david-zlai (approved):

cloud_azure/src/main/scala/ai/chronon/integrations/cloud_azure/Snowflake.scala
```diff
@@ -0,0 +1,158 @@
...
+      }
+      .toList
+      .sortBy(_(partitionColumn))
+
+    snowflakeLogger.info(s"Found ${partitions.size} distinct partitions for table $tableName")
+    partitions
+  }
+
+  private def buildPartitionQuery(tableName: String, partitionColumn: String, partitionFilters: String): String = {
+    val baseQuery = s"SELECT DISTINCT $partitionColumn FROM $tableName"
```
cristian-zlai: I wasn't sure if we had landed on assuming continuity and just checking min and max

cristian-zlai (approved):
Since the snowflake class has no dependencies....
Should not just put it in spark or something?
(I understand it requires the dependencies at runtime for it to work, but as it is now is conveniently stand alone and possible to share with multiple clouds)

Just something to consider.

tchow-zlai:
> Should not just put it in spark or something?



 I had it in spark but since we're doing the AzureKeyVault thing it fails to compile. I didn't bother creating an abstraction for that just yet

PR #1416, opened 2026-02-04
fix: harden stats fetching-- tiles schemas + aggregation type changes

cristian-zlai:
## Summary

Updated context for enhanced stats.

Fetching changes:
* Rather than one aggregator fetching all the IRs for stats -> Fetch each IR individually (helps find issues significantly faster)
* Rather than on set of valueSchema defining the tiles -> If a tile can't be decoded or has extra data an error is logged (this can happen when a join schema is overwritten and a long range is fetched)

Removed straggler from separate PR. 

Added a build_and_start in docker/fetcher for gcp fetcher for faster debugging (requires higher latency acceptance when querying big table)

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [x] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Per-column resilience in stats fetching allowing partial results when some columns fail, plus verbose diagnostics for troubleshooting.

* **Improvements**
  * Per-tile error handling that skips corrupted tiles and reports processed/skipped counts.
  * Local tooling and a debug Docker compose profile to build/start the Fetcher with debug ports and tuned runtime settings.
  * Enhanced logging and diagnostics for stats fetching and tile processing.

* **Documentation**
  * Detailed three-layer hardening guide and debugging workflows added.

* **Chores**
  * Removed obsolete manual staging query submission script.


cristian-zlai:
I want to start with a local end to end tests that replicates the schema evolution.
Then work backwards from there. 
Moved to: crf-add-evolution-test

PR #1434, opened 2026-02-10
feat: Semantic hash in the table location ZIP-1220

tchow-zlai:
## Summary

- create a table named <table_name>_<semantic_hash>
- rename the table to <table_name> 

- Plan to build on top of https://github.com/zipline-ai/chronon/pull/1444 

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Optional semantic-hash support threaded through joins and table writes.
  * Configurable write-format detection; format-specific type identifiers added (Delta, Iceberg).

* **Chores**
  * Explicit "table creation unsupported" behavior for some cloud formats.
  * Write path now defaults file format to PARQUET when creating tables.

* **Tests**
  * Removed a cloud Hudi integration test suite.


cristian-zlai (commented):
As a less surface of changes alternative... would it make sense to put createTable as part of the FormatProvider trait?
Then we have a default method that does the current standard, and for the specific formatProvider (anything with rest I guess) we can do the semantic hash location w/ rename?

cristian-zlai:
Mostly concerned with doing data movement unnecessarily (for rest catalogs seems like there's no data movement on the rename but others may not behave the same).

tchow-zlai:
> Mostly concerned with doing data movement unnecessarily (for rest catalogs seems like there's no data movement on the rename but others may not behave the same).

Actually I believe if it's iceberg tables, the name and location are always decoupled regardless of catalog impl. We are always going to be writing iceberg so it might be a safe assumption to make. 

I'll see about abstracting this somewhere else though, I was hoping we could get rid of our Formatprovider abstraction eventually.

cristian-zlai (approved):

PR #1449, opened 2026-02-12
Fetcher updates - configurable DynDB configs, context class loader etc

piyush-zlai:
## Summary
Cover a couple of fetcher related fixes that were brought up in some conversations:
* Logging additional details when the fetcher handler can't parse the incoming req
* Allowing users to configure DynDB timeouts / configs
* Setting the context class loader in the instrumented tp (depending on event sequencing, we can hit: `Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.ClassLoader.getResourceAsStream(String)" because the return value of "java.lang.Thread.getContextClassLoader()" is null` without it)
* Setting the logResponse in the AwsApi to empty rather than unimplemented as it results in failures. 

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * DynamoDB client now supports configurable concurrency, connection and API call timeouts via environment/configuration.
  * Added region and endpoint override options for DynamoDB with clearer validation and logging.

* **Bug Fixes**
  * Improved thread initialization to ensure third-party libraries (e.g., Spark) can locate classes.

* **Tests**
  * Updated request-parsing tests to reflect input handling changes.


tchow-zlai (approved):

cristian-zlai (approved):
lgtm

PR #1467, opened 2026-02-19
fix: do not publish the same jars twice

tchow-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Chores**
  * Consolidated release asset attachment into a single workflow step that collects and attaches wheel and JAR artifacts together.
  * Updated workflow ordering so notification, cleanup, and promotion steps run after assets are attached.
  * Kept artifact upload destinations intact while centralizing final attachment to ensure uploads occur only after assets are prepared.


.github/workflows/publish_release.yaml
```diff
@@ -468,6 +468,30 @@ jobs:
             jf rt upload $jar "jars/${VERSION#v}/"
           done
 
+  attach-shared-release-assets:
```
david-zlai: why do we need this?

tchow-zlai: there are jars that are common across `aws` , `gcp`, and `azure` that we want to cut the release for. So we release those shared jars just once in a separate step, and the cloud_specific jar releases we do as cloud steps prior.

david-zlai (approved):

PR #1471, opened 2026-02-19
Switch to NettyNioAsyncHttpClient + add iceberg dep

kumar-zlai:
## Summary

EMR bundles aws-sdk-java-bundle which contains CrtConfigurationUtils but not the native aws-crt classes (HttpMonitoringOptions). Due to parent-first classloading, Spark loads CrtConfigurationUtils from the EMR bundle, which then fails to find HttpMonitoringOptions in the system classloader. Switch to pure-Java NettyNioAsyncHttpClient to avoid this cross-classloader issue entirely.

Also add iceberg-spark-runtime-3.5 as an explicit dependency since it was only available as a compile-time transitive dep from the spark module and not included in the assembly jar.

Verified locally that the changes work and I'm not seeing ClassNotFound error anymore

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Chores**
  * Switched AWS HTTP client implementation and updated AWS module dependencies.
* **New Features**
  * Added Iceberg Spark Runtime (3.5) support in the AWS integration.
* **Bug Fixes**
  * Packaging fixes to preserve correct Avro version metadata in assembled artifacts.


tchow-zlai (approved):

cloud_aws/package.mill
```diff
@@ -9,6 +10,8 @@ object `package` extends Cross[CloudAwsModule](build.Constants.scalaVersions) wi
 }
 
 trait CloudAwsModule extends Cross.Module[String] with build.BaseModule {
+  private val avroVersion = build.Constants.avroVersion
```
tchow-zlai: could u just make sure the avro version in the assembly jar is _actually_ this version?

kumar-zlai: Followed up offline and we are actually pulling in the right avro version transitively and our spark jobs succeed now

PR #1540, opened 2026-03-07
feat: runtime optional Databricks OAuth token fetch on EMR

kumar-zlai:
## Summary

EMR step args have a 10,280 character per-argument limit. With Databricks Unity Catalog integration, the OAuth token embedded in spark conf values pushed args past this limit, causing step submission failures.

This moves Databricks OAuth token resolution from the hub to the EMR cluster at runtime. The EMR step fetches service principal credentials from AWS Secrets Manager and exchanges them for a short-lived OAuth token before spark-submit. This avoids the arg size limit since the token is never part of the step args, and improves security since the token never passes through the hub or gets written to disk.

When DATABRICKS_HOST and DATABRICKS_SECRET_NAME are both set in the node environment, the EMR step prepends a token fetch script before spark-submit. Spark conf values referencing $DATABRICKS_OAUTH_TOKEN expand naturally via shell variable. Non-Databricks setups are unaffected.

## Testing

Verified by running an e2e backfill for our AWS databricks staging query

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Adds Databricks OAuth token fetching for EMR job submissions, integrating token retrieval into job startup and using runtime shell-variable expansion for token usage.

* **Configuration Updates**
  * New DATABRICKS_SECRET_NAME environment variable to point to stored Databricks credentials.
  * Runtime token references switched to shell-variable form for catalog access.

* **Bug Fixes**
  * Validates that Databricks host and secret are set together, failing fast if misconfigured.

* **Tests**
  * Added tests covering token-script behavior and quoting rules.


nikhil-zlai (approved):

kumar-zlai:
In the long term we ideally want to have SecretManager interface to fetch Databricks/Snowflake or other secrets needed for auth and put that into our custom connector which will be a separate jar for these integrations that way they will be cloud agnostic and easier to maintain and reuse logic across clouds. CC @nikhil-zlai

PR #1548, opened 2026-03-11
Include Flink health check fn in EKS Flink submitter

piyush-zlai:
## Summary
Add support for injecting a health check function that can check additional aspects of Flink job health before indicating a job is running / healthy. This brings the Eks flink implementation in line with what we have in place for Dataproc. The health check fn could do things like check for successful checkpoints / other things as deemed by the user. 

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Introduced optional health-check mechanism for Flink jobs to enhance job status monitoring. Health checks can refine a RUNNING status to PENDING if verification fails, while other statuses remain unchanged.

* **Tests**
  * Updated test coverage to validate health-check behavior, including URL propagation and conditional status handling based on health check results.


david-zlai (approved):
chatted offline. would like to add a follow up pr in Zipline control plane to timeout if in pending for like 30m or so. 

Would be nice to find a way to surface that somehow though. right now this code just returns jobstatustype. if it's in pending "hell" we won't really know. would have to look at logs

PR #1562, opened 2026-03-13
Fixing endTs bug in metrics loading

varant-zlai:
## Summary
Fix bug that occurs when you fetch metrics with an endTs that overshoots latest existing data.

## Checklist
Did not add unit tests but tested locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed date range generation to prevent queries for future dates that don't exist in the data store.


varant-zlai:
Actually this probably isn't the right approach

PR #1563, opened 2026-03-13
Update FlinkJob to create streaming table prior to kicking off tasks

piyush-zlai:
## Summary
Couple of fixes to get Flink jobs running on AWS:
* Update the user-activities py file to match the logic in GCP (we were missing some casts)
* Issue kvstore.create in the Flink job - this is required for AWS / Dynamo as we create a table per dataset. 
* To ensure the BT flow doesn't break when this is done, I've updated the implementation there to include the admin client by default (this was made optional in the past when we were trying to embed the BT store in a service with a lot of cats threads.. this isn't a concern anymore). 
* If the table exists (like BT / Cosmos) the create call is a no-op

## Testing
Tested by submitting a AWS & GCP Flink job and confirming they both come up and working fine.

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Streaming tables in DynamoDB now auto-configure sort keys.
  * Flink jobs now pre-create streaming datasets before startup.

* **Bug Fixes**
  * Event timestamps normalized to unix_millis(TIMESTAMP(...)) for consistent time handling.

* **Refactor**
  * BigTable integration now requires a concrete admin client and unifies admin client wiring across GCP components.


david-zlai (approved):
just one question. rest looks fine

PR #1575, opened 2026-03-14
fix: ensure flink triggers emit final complete tile on window close

bagipriyank:
## Summary

This PR fixes an issue where Flink window triggers could skip emitting a final tile with `isComplete=true` when windows close, and adds a configurable `allowed_lateness_seconds` property.

## Problem

When using `BufferedProcessingTimeTrigger`, if the processing-time timer fires before the window closes (clearing internal state to `null`), the subsequent `onEventTime()` callback would return `CONTINUE` instead of `FIRE`. This means when Flink's cleanup timer fires at `window.maxTimestamp()`, no final tile with `isComplete=true` would be emitted.

**Impact:**
- Tiles may never be marked as complete
- Client-side caching may not work correctly
- Potential for repeated tile fetches/computations or stale data

### 1. Trigger Fix
Both `AlwaysFireOnElementTrigger` and `BufferedProcessingTimeTrigger` now check if `timestamp >= window.maxTimestamp()` in `onEventTime()` and return `FIRE` to guarantee final tile emission:

```scala
override def onEventTime(timestamp: Long, window: TimeWindow, ctx: TriggerContext): TriggerResult = {
  if (timestamp >= window.maxTimestamp()) {
    TriggerResult.FIRE  // Always emit final complete tile
  } else {
    // existing logic
  }
}
```

### 2. Added support for configuring allowedLateness on windows, which keeps window state open after the watermark passes the window end.

Usage:
```scala
// Via props
Map("allowed_lateness_seconds" -> "60")  // 1 minute
// Via topicInfo
kafka://my-topic/allowed_lateness_seconds=60
```
Default: 0 (disabled)

## Test Plan

- Added unit tests for trigger behavior on window close
- Added unit tests for `allowed_lateness_seconds` configuration parsing
- Added test for the specific bug scenario (processing-time timer fires → clears state → cleanup timer fires)
- Verified triggers fire correctly with allowedLateness configured

## Summary by CodeRabbit

* **New Features**
  * Optional allowed-lateness support for windowed aggregations (configurable in seconds, defaults to 0/disabled).

* **Bug Fixes**
  * Triggers now emit final aggregation when a window closes and properly clean up timers/state.
  * Integration tests updated to expect window-close emissions.

* **Tests**
  * Added comprehensive tests for lateness config parsing/validation and trigger behaviors/edge cases.


piyush-zlai (commented):
Thanks for putting this fix up. 
Largely lgtm - couple of small comments / suggestions.

piyush-zlai (approved):

PR #1584, opened 2026-03-16
Make spark data catalog configurations configurable and one source of truth

david-zlai:
## Summary
^^^

This PR wraps the minimum sparks confs needed to support an input data catalog. Basically a big dictionary wrapper around Glue, Unity, BigQuery, OpenCatalog. 

We store the spark confs that are default values as well as expect users to set the required values that need to come from themselves. 

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Adds reusable cloud catalog presets for AWS, GCP, Azure and Databricks to simplify catalog configuration.

* **Bug Fixes & Improvements**
  * Removes legacy Iceberg/catalog and explicit table-prefix settings; introduces placeholder prompts for user-provided values and simplifies Spark config surface.

* **Refactor**
  * Consolidates inline catalog wiring into modular configuration presets for clearer, maintainable setups.

* **Tests**
  * Adds unit tests for configuration presets and template resolution.


python/src/ai/chronon/repo/spark_catalog_confs.py
```diff
@@ -0,0 +1,92 @@
+class Configuration(dict):
```
tchow-zlai: This is a great idea, but I think it would be better if we actually provided documentation links to the various cloud providers to describe how to set this up. Each provider environment has a different way of doing these things and it also depends on how the user's existing warehouse is set up so it's not super well defined. The teams.py in resources was not _meant_ to work out of the box, it was meant to be a starting point where the user can fill out a config setting that works for them. We can provide the documentation within the scaffolding so that they know where to go. wdyt

tchow-zlai (approved):

PR #1604, opened 2026-03-20
Make EKS Flink job status checks a bit more fine-grained

piyush-zlai:
## Summary
Our EKS Flink jobs can get stuck in Pending state even though they are unhealthy for scenarios such as init container issues (e.g. incorrect s3 paths) OR Flink job start up issues (e.g JM isn't able to get task managers allotted as the ask doesn't fit the shape of what the cluster can support). In these cases, we'd like to not be stuck in pending forever but instead, catch and mark these as failed after a certain amount of time so that the caller upstream can fail / cleanup and potentially retry. 
This PR makes some updates to help:
* If the k8s deployment is 'deployed' and the job manager deploy status is 'error' we fail (e.g. crash looping init container)
* If the k8s deployment is 'deployed' for > timeout (10m) we fail (e.g. unable to provision TM pods) 

## Testing
Covered a couple of failure scenarios and the happy path:
* Incorrect `FLINK_JARS_URI` path - confirmed the status is set to failed
* Very large TM memory size ask for EKS cluster - confirmed status transitions to failed after timeout
* Good config - things work as expected

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate Flink deployment status reporting by considering additional deployment fields and job manager state
  * Automatic failure marking for deployments stuck in non-stable states beyond 10 minutes
  * Improved handling of missing or erroring job manager components to reduce false-positive pending states

* **Tests**
  * Added unit tests covering deployment status resolution across lifecycle, job manager states, and age scenarios


tchow-zlai (approved):

cloud_aws/src/main/scala/ai/chronon/integrations/aws/EksFlinkSubmitter.scala
```diff
@@ -147,23 +151,68 @@ class EksFlinkSubmitter(k8sConfig: Option[Config] = None, ingressBaseUrl: Option
...
+            s"FlinkDeployment $deploymentName has jobManagerDeploymentStatus=MISSING after $DeploymentPendingTimeout, treating as FAILED")
+          JobStatusType.FAILED
+        } else if (timedOut) {
+          logger.warn(
+            s"FlinkDeployment $deploymentName has been in $lifecycleState for more than $DeploymentPendingTimeout without reaching STABLE, treating as FAILED")
+          JobStatusType.FAILED
+        } else {
+          JobStatusType.PENDING
+        }
+      case _ => JobStatusType.UNKNOWN
```
david-zlai: can we log what the case here is? 

for follow up but it would be nice to attach some additional details of why this is UNKNOWN

david-zlai (approved):
one comment around adding logging

PR #1627, opened 2026-03-25
Add MegaTile streaming pipeline: 2 point gets instead of prefix scan

nikhil-zlai:
## Why

Current tiling writes many small hop-aligned tiles per entity (up to 288/day at 5-min resolution). The fetcher does a prefix scan over all tiles and merges them per-window. This causes **read amplification** and serving latency at scale.

## Approach

Replace the prefix-scan tile scheme with **2 point gets per entity**: `(key, today)` + `(key, yesterday)` from the stream KV store, plus `(key)` from the batch KV store.

### Flink (write side)
- **MegaTileStreamProcessor**: per-entity state manager maintaining small-window tiles (sawtooth running IR corrected by periodic eviction) and large-window per-day accumulators
- Small windows (≤ 2d): tiles per tier + `cachedSmallWindowIr` (rebuilt from tiles on eviction via `buildMegaTileIr`)
- Large windows (> 2d): daily running IR (today + yesterday), routed by event timestamp
- Day transitions driven by watermark (not event timestamps) to prevent future-timestamped events from corrupting state
- Late events (up to 2d) correctly routed to yesterday's accumulator; multi-day watermark jumps clear stale state
- Single packed `Array[Any]` per day: small-window columns + large-window columns emitted to KV store
- **MegaTileProcessFunction**: thin Flink `KeyedProcessFunction` wrapper handling state serde (via `MegaTileCodec.encodeBaseIr`/`decodeBaseIr` for tiles, `encode`/`decode` for windowed IRs), event-time timers for eviction, and key-switch reinitialization
- **MegaTileAvroCodecFn**: packs mega tile output into KV `PutRequest` with `TileKey(dataset, entityKey, DayMillis, dayStart)`
- Shared `buildTiledTail`/`buildMegaTiledTail` in `BaseFlinkJob` — both `FlinkGroupByStreamingJob` and `ChainedGroupByJob` use the same pipeline tail

### Fetcher (read side)
- **MegaTileMerger**: per-column merge of `(today, yesterday, batchIr)` → finalized result
  - Small windows: pick today's entry (self-contained). Fall back to yesterday only if today's entry is entirely absent from KV.
  - Large windows + unwindowed: batch collapsed + streaming daily aggregates (sum yesterday if batch stale) + tail hops
- **MegaTileCodec**: Avro serde for windowed mega tile IRs + base (unwindowed) tile IRs
- **GroupByFetcher**: constructs 2 point-get `GetRequest`s (today + yesterday) when `isMegaTilingEnabled`; resolves `queryTs` once to avoid midnight-boundary inconsistency
- **GroupByResponseHandler**: `mergeMegaTilesFromStreaming` branch using `getBatchIrFromBatchResponse` (handles cached batch correctly)

### Opt-in
- `OnlineStrategy` thrift enum (`DEFAULT`, `STREAMING_MEGATILES`) on `GroupBy` field 8
- `GroupByOps.isMegaTilingEnabled` checks the enum; excluded from `semanticHash` (online-only, doesn't trigger batch recomputation)
- Both Flink and fetcher check the same field

### Boundary fix (original commit)
- `alignedCollapsedBoundary` in `SawtoothMutationAggregator` aligns the collapsed/tail split to hop boundaries, preventing double-counting when shared tail hops span windows with different collapsed boundaries

## Test plan

**3 layers of fuzz tests, all comparing against `NaiveAggregator`:**

- [x] `MegaTileAggregatorTest` (6 tests) — tile building + `serveMegaTile` merge: batch fresh, batch delayed, idle entities, eviction boundaries, batchEnd boundary, multiple agg types (SUM/COUNT/AVG/MIN/MAX/LAST/FIRST)
- [x] `MegaTileMergerTest` (5 tests) — per-day entry split + `MegaTileMerger.merge`: batch fresh, batch delayed, idle entities, multiple agg types, day boundary transitions
- [x] `MegaTileStreamProcessorTest` (6 tests) — full Flink simulation with sawtooth running IR + eviction at tile boundaries: batch fresh, batch delayed, idle entities, all agg types, day boundaries, **multi-day watermark gap** (3-day idle then resume)
- [x] `MegaTileCodecRoundTripTest` (3 tests) — serde round-trips every 50 events through `MegaTileCodec.encode`/`decode` + `encodeBaseIr`/`decodeBaseIr`: batch fresh, all agg types, **key switch verification** (Flink key reuse)

Windows tested: `6h, 1d, 47h, 2d, 49h, 3d, 7d`

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
@@ -129,13 +129,14 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
       if (window != null && window.millis > tailBufferMillis) {
         val hopIndex = tailHopIndices(i)
         val queryTail = TsUtils.round(queryTs - windowMillis, hopSizes(hopIndex))
+        val alignedCollapsed = alignedCollapsedBoundary(batchEndTs - windowMillis, i)
         val hopIrs = batchIr.tailHops(hopIndex)
         val relevantHops = mutable.ArrayBuffer[Any](ir(i))
         var idx: Int = 0
         while (idx < hopIrs.length) {
           val hopIr = hopIrs(idx)
           val hopStart = hopIr.last.asInstanceOf[Long]
-          if ((batchEndTs - windowMillis) + tailBufferMillis > hopStart && hopStart >= queryTail) {
+          if (alignedCollapsed > hopStart && hopStart >= queryTail) {
             relevantHops += hopIr(baseIrIndices(i))
           }
           idx += 1
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
@@ -118,7 +118,7 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
   }
 
   // Like mergeTailHops but skips NO BATCH columns (window <= tailBuffer)
-  private def mergeTailHopsForBatchColumns(ir: Array[Any],
+  private[windowing] def mergeTailHopsForBatchColumns(ir: Array[Any],
                                            queryTs: Long,
                                            batchEndTs: Long,
                                            batchIr: FinalBatchIr): Array[Any] = {
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
@@ -133,7 +133,13 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
           val hopIr = hopIrs(idx)
           val hopStart = hopIr.last.asInstanceOf[Long]
           if (alignedCollapsed > hopStart && hopStart >= queryTail) {
-            relevantHops += hopIr(baseIrIndices(i))
+            val hopVal = hopIr(baseIrIndices(i))
+            if (relevantHops(0) == null) {
+              // bulkMerge may mutate its first non-null IR; clone cached tail-hop state on that path.
+              relevantHops += windowedAggregator(i).clone(hopVal)
+            } else {
+              relevantHops += hopVal
+            }
           }
           idx += 1
         }
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
@@ -65,9 +65,7 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
   }
 
   // Build windowed mega tile IR from small tiles
-  def buildMegaTileIr(tiles: Map[Long, collection.Map[Long, Array[Any]]],
-                      now: Long,
-                      batchEnd: Long): Array[Any] = {
+  def buildMegaTileIr(tiles: Map[Long, collection.Map[Long, Array[Any]]], now: Long, batchEnd: Long): Array[Any] = {
     val megaTileIr = windowedAggregator.init
     var col = 0
     while (col < windowedAggregator.length) {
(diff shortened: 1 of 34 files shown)


aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
```diff
@@ -0,0 +1,151 @@
...
+      batchEnd
+    }
+  }
+
+  // Compute tile start timestamps for an event across all active tiers
+  def tileStartsForEvent(eventTs: Long): Map[Long, Long] =
+    activeTiers.iterator.map(hopSize => hopSize -> TsUtils.round(eventTs, hopSize)).toMap
+
+  // Per-tier retention floor: the earliest tile we must keep
+  def retentionFloor(hopSize: Long, now: Long, batchEnd: Long): Long = {
```
nikhil-zlai: rename `streamingRetentionFloor`

piyush-zlai (commented):
Had some qs / comments around some of the event triggers stuff, thought it might be good to bounce off. 

Given this is a fairly large & exhaustive PR - it would be good to lock down our confidence that it won't a) break existing folks that don't opt in b) if they do opt in that there are no major issues. 

It would be swell to test in our canary with a couple of jobs with flag on / off. For the flag off case - it might be worth testing the restore from savepoint behavior (as I think we did seemingly refactor the flink job code (worry is around uids lining up, if they dont we can't restore from savepoint).

flink/src/main/scala/ai/chronon/flink/window/MegaTileProcessFunction.scala
```diff
@@ -0,0 +1,288 @@
...
+      if (result.todayEntry != null) {
+        out.collect(
+          new TimestampedTile(ctx.getCurrentKey,
+                              megaTileCodec.encode(result.todayEntry),
+                              result.todayStart,
+                              System.currentTimeMillis()))
+      }
+
+      if (processor.hasSmallWindows) {
+        ctx.timerService().registerEventTimeTimer(timestamp + processor.minSmallWindowTileSize)
```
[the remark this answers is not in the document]
nikhil-zlai: not all keys need the same timer? only the ones that need eviction at that point no?
my understanding is that flink timers are cheap.

flink/src/main/scala/ai/chronon/flink/window/MegaTileProcessFunction.scala
```diff
@@ -0,0 +1,335 @@
...
+      if (result.yesterdayEntry != null) {
+        out.collect(
+          new TimestampedTile(keys,
+                              megaTileCodec.encode(result.yesterdayEntry),
+                              result.yesterdayStart,
+                              event.startProcessingTimeMillis))
+      }
+
+      // Schedule eviction timer if small windows exist
+      if (processor.hasSmallWindows) {
```
abbywh: I think scheduling timers for small windows makes a ton of sense. This is a pita to use a KV store TTL on. But these should be pretty reasonable to jitter over as well, IMO, similar to @piyush-zlai 's comment below. The penalty for a timer that is 0-5 minutes late here is only one tile overfetch.

flink/src/main/scala/ai/chronon/flink/window/MegaTileProcessFunction.scala
```diff
@@ -0,0 +1,335 @@
...
+        val hopSize = processor.minSmallWindowTileSize
+        val watermark = ctx.timerService().currentWatermark()
+        val procNow = ctx.timerService().currentProcessingTime()
+
+        if (!isLive && isCaughtUp(watermark, procNow)) {
```
abbywh: Similar to the note about `onTimer`, can we just skip in catch up mode internal to the `megaTileStreamProcesser`? 

You should be able to `MegaTileState.apply(currentState, currentWatermark, currentTimer)` and handle eviction there.

piyush-zlai (approved):
One idea to make this safer might be to carve out the refactors of the existing jobs as a separate PR that we review + merge and then rebase this to just be the net new. 

Currently feels like there's a lot of intermixing of the two (e.g. in the chained / regular flink job code) so its a bit hard to tease apart how risky these are to merge for existing users..

PR #1641, opened 2026-03-28
CLI: Add support for multiple conf paths for `eval`

NDR-114:
## Summary
This PR aims to allow multiple conf paths for an `eval`, mainly for easier use in CI/pre-commit hooks.

Changes:

- eval now takes one or more CONFS arguments instead of a single CONF
- Hub instantiation and diff uploads are deduplicated per hub URL; if multiple `confs` share the same hub, it only connects and uploads once
- Auth/info log lines are suppressed for subsequent confs on the same hub
- Text output labels each result with the conf path when multiple confs are provided
- JSON output wraps results in an aggregated {"status": ..., "results": [...]} envelope for multi-conf runs; single-conf output is unchanged
- Added tests

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

### DIFF
--- a/python/src/ai/chronon/repo/hub_runner.py
+++ b/python/src/ai/chronon/repo/hub_runner.py
@@ -14,7 +14,7 @@ from ai.chronon.cli.formatter import (
     format_print,
     jsonify_exceptions_if_json_format,
 )
-from ai.chronon.cli.git_utils import get_current_branch
+from ai.chronon.cli.git_utils import get_current_branch, get_git_user_email
 from ai.chronon.cli.theme import (
     print_error,
     print_info,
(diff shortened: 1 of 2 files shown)


cristian-zlai:
Eval when used in CI uses a special parameter tempCatalogOverride and warehouseOverride that signal eval two things:
* Do not clean up the catalog where the tables are being written. (helps for running additional checks post creation)
* This is the working space, so use whatever you see there if it's available. 

In this context you can build incrementally on a CI catalog by doing sequential eval requests and there won't be any wasted compute. (CI is in charge of ensuring a cleanup of the catalog once the job is done). 

The main concern with adding multiple confs becomes that conf parameter is already used in almost every command (run-adhoc, schedule, backfill), giving it a different semantic for eval would add to the cognitive load for developers to know that in the case of eval its different.

If we still think valuable adding multiple confs (in which case they'd just evaluate sequentially) I'd recommend adding an optional parameter (`--additional-confs`) which gets passed in the params to the `eval` request. (additionalConfs) we can add the logic for that in eval module. (eval intentionally tries to keep all the logic within the scala module to reduce integration testing)

PR #1646, opened 2026-03-29
fix source types for list invariance

BenMagyar:
## Summary

Python lists are invariant, so if we have a case of `List[EventSource]` that does not  meet the requirement for `list[EventSource | EntitySource | etc]`. `Sequence` is covariant so it would match. 

Swapping from `List` to `Sequence` and then normalizing down to the `List[api.Source]` type post-call. Added a check against the expanded `Tuple` case as well. 

## Checklist
- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Source parameters now accept any sequence type (e.g., tuples) in addition to lists.

* **Bug Fixes**
  * Centralized source normalization preserves iteration order and handles single vs. multiple sources consistently.
  * Validation now reports indexed errors for invalid sequence elements.

* **Tests**
  * Added tests verifying sequence normalization, ordering, and invalid-element error reporting.

### DIFF
--- a/python/src/ai/chronon/group_by.py
+++ b/python/src/ai/chronon/group_by.py
@@ -15,6 +15,7 @@
 import inspect
 import json
 import logging
+from collections.abc import Sequence
 from copy import deepcopy
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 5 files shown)


tchow-zlai (approved):

nikhil-zlai (approved):
TIL

PR #1673, opened 2026-04-01
Include chronon-fetcher image in zipline install command

BenMagyar:
## Summary

When `zipline install` runs it should include the image necessary for the fetcher `chronon-fetcher`. No existing covering tests on which images run (its just a list) so skipped.

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Release Notes

* **New Features**
  * Added fetcher application support to deployments.

* **Chores**
  * Updated deployment summary output to dynamically list all application images.
  * Updated registry configuration to include fetcher image reference.

### DIFF
--- a/python/src/ai/chronon/repo/admin.py
+++ b/python/src/ai/chronon/repo/admin.py
@@ -60,6 +60,7 @@ def _app_images(cloud):
         ("hub", f"ziplineai/hub-{cloud}"),
         ("eval", f"ziplineai/eval-{cloud}"),
         ("frontend", "ziplineai/web-ui"),
+        ("fetcher", "ziplineai/chronon-fetcher"),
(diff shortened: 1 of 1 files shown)


piyush-zlai (approved):

tchow-zlai (approved):

tchow-zlai:
@BenMagyar fwiw I believe the fetcher is not gated behind an enterprise API token so you could technically just pull it down

BenMagyar:
@tchow-zlai Yeah we do that now, but if we're running an install workflow it means we either have the pull + push for some vs the passthrough for others which feels a bit awkward vs just one pull + push path.

PR #1680, opened 2026-04-02
[codex] Port MegaTile streaming and serving fixes

mickjermsurawong-openai:
## Behavior change over the base MegaTile PR

This PR keeps the base MegaTile serving model, but tightens the edge-case behavior around late data, day rollover, and Flink catchup/live transitions.

The main serving effect is that MegaTile reads now fetch and merge the relevant daily streaming rows from the query day back to the batch day, including fallback-day coverage. When no stream row exists for a day, NO BATCH MegaTile outputs are cleared instead of leaving stale values behind.

The main streaming effect is that event time still owns retained tile and day routing, while a separate small-window as-of timestamp controls whether a retained late event can update the cached no-batch small-window row. This allows retained late events to update 1d/large-window state without re-inflating stale 1h values outside the current serving horizon.

The Flink MegaTile path now chooses behavior explicitly by mode: no watermark, active catchup, sparse-key lag, or live. Active catchup follows the watermark frontier, sparse/live paths use processing time for materialized-view decay, events older than yesterday are dropped before mutating state, and rollover emits use logical UTC day-start keys.

Buffered MegaTile emits also support optional stable per-key jitter through `buffering_output_jitter_millis`; the default remains disabled, so existing behavior is unchanged unless the property is set.

## Implementation notes

- Fetch and merge historical MegaTile daily stream rows for serving, with no-batch null clearing and per-thread safe MegaTile codec decode.
- Rebuild MegaTile small-window eviction state in one pass, and avoid mutating cached batch tail-hop IR during repeated serves.
- Add `smallWindowAsOfTs` to stream ingestion so retained base tiles are separate from cached small-window visibility.
- Add Flink catchup/live mode selection, watermark-lag slack, logical day-start emits, stale-event admission checks, and optional buffered emit jitter.
- Wire `buffering_output_time_millis` and `buffering_output_jitter_millis` through single and chained MegaTile Flink jobs.

## Tests

- Added stream-processor and codec regressions for retained late events updating 1d without stale 1h inflation.
- Added a Flink process-function harness covering live/catchup boundary behavior, UTC day rollover, upstream stop/resume, consumer lag, sparse-key fallback, cold-start backlog, stale event drops, checkpoint restore, malformed timestamps, emit/evict timer collision, and stable jitter.
- Extended Flink job integration coverage for stale UTC-midnight watermark behavior and mixed small/large-window daily snapshots.

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
@@ -87,13 +87,7 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
     megaTileIr
   }
 
-  // Single-entry merge helper used by MegaTileAggregatorTest for retained-window simulations.
-  // Here megaTileIr is already built from enough retained tiles to cover each small window,
-  // including pre-batchEnd tiles when the small window crosses batchEnd. That makes small-window
-  // columns self-contained, so adding batch tail below would double-count them.
-  //
-  // Online per-day KV serving uses MegaTileMerger instead. There the daily row owns only the
-  // stream-side slice, so MegaTileMerger must add batch tail for crossing small windows.
+  // Merge mega tile with batch IR to produce finalized result
   def serveMegaTile(batchIr: FinalBatchIr, megaTileIr: Array[Any], queryTs: Long, batchEnd: Long): Array[Any] = {
     // Start from batch collapsed (normalized) — same as lambdaAggregateIrTiled
     val resultIr = if (batchIr != null) windowedAggregator.clone(batchIr.collapsed) else windowedAggregator.init
@@ -101,7 +95,7 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
     while (col < windowedAggregator.length) {
       val window = windowMappings(col).aggregationPart.window
       if (window != null && window.millis <= tailBufferMillis) {
-        // Small window: megaTileIr already covers the full retained-window range for this query.
+        // NO BATCH: mega tile is self-contained, replace batch value
         resultIr(col) = megaTileIr(col)
       } else if (megaTileIr(col) != null) {
         // BATCH or unwindowed: merge mega tile into batch collapsed
@@ -110,7 +104,8 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
       col += 1
     }
 
-    // Tail hops only for BATCH columns. Small windows are self-contained in this retained-window helper.
+    // Tail hops ONLY for BATCH columns (window > tailBuffer).
+    // NO BATCH columns are fully covered by the mega tile — adding tail hops would double-count.
     if (batchIr != null) {
       mergeTailHopsForBatchColumns(resultIr, queryTs, batchEnd, batchIr)
     }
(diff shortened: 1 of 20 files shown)


aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileAggregator.scala
```diff
@@ -133,7 +133,13 @@ class MegaTileAggregator(aggregations: Seq[Aggregation],
...
           val hopStart = hopIr.last.asInstanceOf[Long]
           if (alignedCollapsed > hopStart && hopStart >= queryTail) {
-            relevantHops += hopIr(baseIrIndices(i))
+            val hopVal = hopIr(baseIrIndices(i))
+            if (relevantHops(0) == null) {
+              // bulkMerge may mutate its first non-null IR; clone cached tail-hop state on that path.
+              relevantHops += windowedAggregator(i).clone(hopVal)
+            } else {
+              relevantHops += hopVal
+            }
```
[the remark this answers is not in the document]
nikhil-zlai: nice! good catch!

aggregator/src/main/scala/ai/chronon/aggregator/windowing/MegaTileStreamProcessor.scala
```diff
@@ -63,23 +66,38 @@ class MegaTileStreamProcessor(val megaTileAgg: MegaTileAggregator, val store: Ti
...
+      if (acceptedTileStarts.nonEmpty) {
+        var cachedIr: Array[Any] = null
+        var cachedIrUpdated = false
         var col = 0
         while (col < windowedAgg.length) {
           if (isNoBatch(col)) {
-            windowedAgg.columnAggregators(col).update(cachedIr, row)
+            val hopSize = columnHopSize(col)
+            acceptedTileStarts.get(hopSize).foreach { tileStart =>
+              val effectiveStart = megaTileAgg.effectiveStart(col, smallWindowAsOfTs, currentDayStart)
```
mickjermsurawong-openai: Follow-up ported here: `smallWindowAsOfTs` bounds only the cached no-batch small-window IR. Event time still owns retained base-tile/day routing, so retained late rows can update 1d without re-inflating stale 1h.

flink/src/main/scala/ai/chronon/flink/window/MegaTileProcessFunction.scala
```diff
@@ -187,40 +215,260 @@ class MegaTileProcessFunction(
...
+
+  private case object NoWatermark extends Mode
+
+  private case object ActiveCatchup extends Mode
+
+  private case object SparseKeyLag extends Mode
+
+  private case object Live extends Mode
+
+  private def currentMode(processingTs: Long, watermark: Long): Mode = {
```
mickjermsurawong-openai: Follow-up ported here: the mode split decides both eviction time and event small-window as-of time. `ActiveCatchup` follows the watermark frontier, while sparse/live/no-watermark paths preserve PT/event-time behavior as appropriate; buffered jitter stays opt-in via config.

flink/src/main/scala/ai/chronon/flink/FlinkGroupByStreamingJob.scala
```diff
@@ -126,7 +130,14 @@ class FlinkGroupByStreamingJob(eventSrc: FlinkSource[ProjectedEvent],
...
     logger.info(f"Running Mega Tiled Flink job for groupByName=${groupByName}, Topic=${topic}.")
     val preparedStream = buildSourceStream(env)
-    buildMegaTiledTail(preparedStream, inputSchema, parallelism, sinkFn, kvStoreCapacity, enableDebug)
+    buildMegaTiledTail(preparedStream,
+                       inputSchema,
+                       parallelism,
+                       sinkFn,
+                       kvStoreCapacity,
+                       enableDebug,
+                       bufferingOutputTimeMillis,
```
nikhil-zlai: nice! thanks for adding this!

nikhil-zlai (approved):

PR #1687, opened 2026-04-03
fix: Recursive namespace propagation

tchow-zlai:
Replaces the shallow, type-specific namespace propagation in `update_metadata` with a single recursive
  `_propagate_namespace` function that walks the entire config tree.

  ### Problem

  When a Join uses a GroupBy that reads from another Join via `JoinSource` (chained joins), the compile step
  only propagated `outputNamespace` one level deep. In a chain like:

  Join A (4 groupbys) → GroupBy_v1 (reads from Join A via JoinSource) → Join B (uses GroupBy_v1)

  Compiling Join B would set namespace on `GroupBy_v1` but never recurse into its `JoinSource` to reach Join
  A. At runtime, Join A's output table resolved to `null.tablename` instead of `warehouse_db.tablename`.

  ### Fix

  The old code had three separate, non-recursive helpers (`_set_part_metadata`,
  `_propagate_namespace_to_join_sources`, and type-specific `if/elif` branches). These are replaced by a
  single `_propagate_namespace(node)` that:

  - Sets `outputNamespace`, `team`, and merges execution info on any node with `metaData` (`GroupBy`, `Join`,
  `Model`, `ModelTransforms`, `StagingQuery`)
  - Creates `MetaData()` on nodes that lack it (restoring pre-refactor safety)
  - Recurses into `joinParts` (Joins), `models` (ModelTransforms), `sources` (any node), and `Join.left`
  - Preserves explicitly-set namespaces (won't overwrite)

  All five config types are handled uniformly, eliminating the redundant second `merge_team_execution_info`
  call that previously existed for the top-level object.

  ### Tests

  6 new test cases covering:
  - Join → GroupBy with JoinSource → inner Join (the reported bug)
  - Arbitrary depth chaining (Join → GB → Join → GB → Join)
  - Join with `left` as a JoinSource
  - Explicit namespace preservation in mid-chain nodes
  - Standalone GroupBy with JoinSource
  - Full reproduction of the user-reported topology (4 GBs → Join → GB_v1 → second Join)

  ### Compiled config changes

  Recompiled canary configs now include `"common": {}` on `clusterConf` for nodes that previously omitted it.
  This is a no-op — an empty map is semantically identical to the field being absent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Metadata (outputNamespace, team, executionInfo) now reliably propagates into nested joins, group-bys and join-sources; nested executionInfo/conf/env blocks were added where needed and useLongNames/outputNamespace flags ensured.
  * Increased cluster idle timeouts and lifecycle TTLs (300 → 7200) and reduced online/offline stepDays (30 → 10) in affected configs.

* **Tests**
  * Added unit tests covering deep nested propagation and namespace-preservation scenarios.


nikhil-zlai (approved):

nikhil-zlai:
excellent summary!

PR #1689, opened 2026-04-03
Unify partition handling with MAX-based readiness checks

nikhil-zlai:
## Summary
- Replace 4 separate partition sensing code paths (dense Hive, time-partitioned, trigger-based, unpartitioned) with a single unified approach: `lastAvailablePartition >= range.end`
- For string partition columns (Hive `ds`), `MAX(col)` is the ready date. For timestamp/date columns, `DATE(MAX(col)) - 1 day` — seeing data on day N proves N-1 is complete
- Deprecate `timePartitioned` flag — column type is now detected automatically from the schema
- Support sparse partitions: gaps in partition ranges no longer cause sensor failures

## Key changes
- **Format trait**: Added `lastAvailablePartition()` and `firstAvailablePartition()` that inspect column type and apply the right logic
- **BatchNodeRunner.checkPartitions()**: Collapsed from 4 cases (~140 lines) to 3 cases (~60 lines): triggerExpr → partitionColumn → table existence
- **BatchNodeRunner.computeInputTablePartitionStatuses()**: Replaced partition enumeration + set diff with `lastAvailablePartition`-based readiness check
- **TableUtils**: `tableCoversRange()`, `lastAvailablePartition()`, `firstAvailablePartition()` now delegate to Format trait
- **TableDependencies**: Stopped propagating `timePartitioned` flag
- **Downstream callers**: Updated StepRunner, StagingQuery, Analyzer, JoinBase, PartitionRunner, MergeJob

## Test plan
- [x] All 26 BatchNodeRunnerTest tests pass (including timestamp column tests without `timePartitioned` flag)
- [x] All 25 TableUtilsTest tests pass
- [x] All 14 TableDependenciesTest tests pass (updated to remove `timePartitioned` assertions)
- [ ] CI runs full test suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Relaxed validation so time-partitioned inputs no longer error when a partition column is unset.

* **New Features**
  * Added first/last available-partition helpers and a hook to resolve partition column names.

* **Improvements**
  * Switched from enumerating all partitions to boundary-based checks for readiness, verification, and range computations.
  * Simplified runtime readiness, join/group-by, and post-run validations to use boundary comparisons.

* **Deprecations**
  * Marked older partition-enumeration utilities as deprecated.

* **Tests**
  * Updated tests to expect boundary-based semantics and adjusted failure messages.


tchow-zlai (commented):
Doing a pass at the API:

[additional_partitions](https://github.com/zipline-ai/chronon/blob/79544ed366212a07a8f34c5050dffea3980f2f2c/python/src/ai/chronon/staging_query.py#L31) I believe will no longer work. We should deprecate.

[time_partitioned](https://github.com/zipline-ai/chronon/blob/79544ed366212a07a8f34c5050dffea3980f2f2c/python/src/ai/chronon/staging_query.py#L33) should be removed as that's the point of this PR

spark/src/main/scala/ai/chronon/spark/batch/BatchNodeRunner.scala
```diff
@@ -75,53 +75,9 @@ class BatchNodeRunner(node: Node, tableUtils: TableUtils, api: Api) extends Node
...
-              new RuntimeException(s"Sensor timed out after ${retryIntervalMin * attempt} minutes. ${e.getMessage}", e))
-        }
-      }
-      return retryMaxCheck(0)
-    }
-
-    // Case 3: No partitions but trigger expression is defined
-    if (!hasPartitionColumn && hasTriggerExpr) {
+    // Case 1: triggerExpr overrides — user-defined readiness check
+    if (hasTriggerExpr) {
```
tchow-zlai: we probably need to introuce `trigger_expr` on `Source`

nikhil-zlai:
> Doing a pass at the API:
> 
> [additional_partitions](https://github.com/zipline-ai/chronon/blob/79544ed366212a07a8f34c5050dffea3980f2f2c/python/src/ai/chronon/staging_query.py#L31) I believe will no longer work. We should deprecate.
> 
> [time_partitioned](https://github.com/zipline-ai/chronon/blob/79544ed366212a07a8f34c5050dffea3980f2f2c/python/src/ai/chronon/staging_query.py#L33) should be removed as that's the point of this PR

@tchow-zlai that was intentional to not break compile of existing customers. at the very least i would like to do it as a follow up.

tchow-zlai (commented):
1. PartitionRunner.scala:65-66 — tu.partitions(inputTable) and tu.partitions(inputTable).filter(_ <= endDs)
  in computeRanges. The PR only changes partitionRange() (line 131) but not computeRanges. This is a genuine
  miss — still enumerates all partitions to compute set diffs for stats jobs.
  2. PartitionRunner.scala:74 — tu.partitions(outputTable) — same, still in computeRanges.
  3. Driver.scala:1045 — tableUtils.partitions(tbl, spec.tail.toMap, ...) — uses subPartitionFilters, which is
   a legitimate use case that can't be replaced by boundary checks.

spark/src/main/scala/ai/chronon/spark/catalog/Format.scala
```diff
@@ -107,8 +107,84 @@ trait Format {
...
 
-  // Returns the max date present in a timestamp/date column, formatted as a partition date string.
-  // Used by sensors to check if data covers a required date range.
+  // Unified last available partition: handles both string partition columns and timestamp/date columns.
+  // For string columns that are catalog partitions (Hive/Iceberg/Delta), uses metadata-only lookup — no data scan.
+  // For timestamp/date columns, falls back to a scan: DATE(MAX(col)) - 1 day.
+  def lastAvailablePartition(tableName: String, partitionColumn: String, partitionSpec: PartitionSpec)(implicit
+      sparkSession: SparkSession): Option[String] = {
+    // Try metadata-based partition listing first (free for Hive/Iceberg/Delta)
+    val metadataResult = Try(primaryPartitions(tableName, partitionColumn, "")(sparkSession)) match {
```
tchow-zlai: The Format trait's default lastAvailablePartition tries primaryPartitions first (metadata-based), then falls
   back to sparkSession.read.table(tableName). For Snowflake tables, both will fail — primaryPartitions
  returns empty (Snowflake's partitions() discovers its own column, as we debugged earlier), and
  sparkSession.read.table() can't read Snowflake tables. The sensor would always fail with "Could not
  determine last available partition."

spark/src/main/scala/ai/chronon/spark/batch/BatchNodeRunner.scala
tchow-zlai: can be a followup

kumar-zlai:
Verified the changes for UC delta/iceberg tables. Got a successful backfill run on sparse input delta table and also able to read SQ iceberg output chronon generates as input to groupBy which was a problem before due to this issue `Unity Catalog supports only liquid clustering and interprets partitions specified in the PARTITION BY clause as clustering keys for liquid clustering.` https://docs.databricks.com/aws/en/tables/partitions. 

Found a couple of issues during e2e testing

1. Join fails when it produces no output due to missing data we probably need to log a warning instead of throwing error https://github.com/zipline-ai/chronon/blob/main/spark/src/main/scala/ai/chronon/spark/batch/BatchNodeRunner.scala#L263 (we can include this in this PR) CC @tchow-zlai 
2. We now let the jobs succeed when there is no data but we overwrite step status with kv partition info thinking output got deleted and are scheduling them again. I can address this in a separate PR and also it's in hub.

tchow-zlai:
We need to test this for Snowflake and BigQuery. Also make sure this works for partitioned tables as well.

cristian-zlai:
Successfully tested on Snowflake (with holes and custom partition name)

cristian-zlai:
Ran bigtable exports [pubsub dump + native]:
https://canary.zipline.ai/stagingqueries/gcp.exports.dim_listings_test__0/offline/status?start=1774483200000&end=1775520000000&isScheduled=false

cristian-zlai (approved):

PR #1693, opened 2026-04-04
fix: Propagate useLongNames to nested Joins and handle null in merge

tchow-zlai:
## Summary
- Fix `useLongNames` not propagating through `GroupBy → JoinSource → inner Join` in chained join configs
- Replace `getattr` with explicit `isinstance` checks for source traversal
- Fix `_merge_mode_maps` to early-return `None` when all inputs are empty, preventing potential `AttributeError`

## Test plan
- [ ] `python -m pytest python/test/test_parse_teams.py` — 16 tests pass
- [ ] `test_use_long_names_propagates_to_nested_join_via_join_source` — verifies chained propagation
- [ ] `test_use_long_names_does_not_override_explicit_false` — verifies explicit values preserved
- [ ] `test_merge_mode_maps_with_all_none_inputs` — verifies null safety

## Summary by CodeRabbit

* **Bug Fixes**
  * Naming-convention propagation now applies only to direct join parts (does not leak into nested join sources).
  * Merge logic now returns no result immediately when all mode inputs are empty, preventing spurious merges.

* **Tests**
  * Added tests covering direct join-part inheritance behavior and merge behavior when all inputs are null.


nikhil-zlai:
should we propagate though? what if another reader of the join wants short names?

i think each join should controls its own naming

nikhil-zlai (approved):

PR #1715, opened 2026-04-06
Attempt auth native Zipline hub way first

david-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


sean-zlai (approved):
Makes sense.  Would be nice to add some comments explaining each logic branch, and some [tests](https://github.com/zipline-ai/chronon/blob/b9907bd2604e73943d9881478e95b1df3b2764c1/python/test/test_zipline_hub.py) to validate this flow if possible.

From reading through this it should still fall through to the older GCP/Azure token flows, as we want to make sure we don't break those use cases until everyone has transitioned.

PR #1743, opened 2026-04-16
Treat empty partition metadata as table is unpartitioned

BenMagyar:
## Summary

In #1689 we routed down this path that previously would be skipped during the forced handling of time partitioned tables. For Databricks Iceberg tables where the table is time partitioned (really clustered over the time col), the partition check grab returns null partition metadata so the `.parts` call fails on a null pointer exception.

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer partition metadata handling: treat absent or empty metadata as missing, returning no partition when metadata is empty and falling back to table scans when needed.
  * Exclude null-valued partition entries when building partition lists to avoid invalid entries and downstream errors.

* **Tests**
  * Added coverage verifying fallback behavior when partition metadata is null and ensuring correct min/max partition results from scans.


nikhil-zlai (approved):
Thanks for the fix and test coverage! 🙏🏾

PR #1747, opened 2026-04-17
feat: Support dynamic overwrites on time columns in unpartitioned tables CTRL-145

tchow-zlai:
Summary

  Fix duplicate data on Iceberg tables without a real partition spec (e.g. Unity Catalog managed tables
  with liquid clustering). insertInto() with partitionOverwriteMode=DYNAMIC silently appends instead of
  replacing on unpartitioned Iceberg tables, creating duplicates on any rewrite.

  For tables with a real Iceberg partition spec, the existing insertInto() path is preserved (works
  correctly with dynamic partition overwrite). For unpartitioned tables, the new path uses MERGE INTO ...
  ON FALSE with WHEN NOT MATCHED BY SOURCE to atomically delete existing rows in the target range and
  insert new rows — all in a single Iceberg snapshot.

  The problem

  Spark's partitionOverwriteMode=DYNAMIC only works when the table has Hive-style or Iceberg partition
  directories. On unpartitioned Iceberg tables (including UC managed tables that convert PARTITION BY to
  liquid clustering), insertInto() appends instead of replacing. This caused silent duplicates on any
  recompute — Yaz hit this when trying to fix a partial backfill.

  The fix

  - Detect unpartitioned Iceberg tables via Iceberg.partitionColumnNames(), which inspects the .partitions
   metadata table for partition spec fields.
  - Atomic range overwrite using MERGE INTO with ON FALSE:
    - WHEN NOT MATCHED BY SOURCE AND target.ds >= min AND target.ds <= max THEN DELETE — deletes all
  existing rows in the range
    - WHEN NOT MATCHED THEN INSERT * — inserts all new rows
    - Single Iceberg snapshot — if the write fails, the table is untouched. Concurrent-writer safe via
  Iceberg's optimistic concurrency.
  - Type-safe range bounds using lit(value).expr.sql — produces correct SQL literals for STRING, DATE,
  TIMESTAMP, INT, etc.
  - ChrononMergeExtension — Spark 3.5's parser doesn't support WHEN NOT MATCHED BY SOURCE (Spark 4.0
  feature), but the execution engine does. This parser extension intercepts the syntax, extracts the
  clause via regex, and constructs the MergeIntoTable plan with notMatchedBySourceActions populated.
  Opt-in via spark.sql.extensions.

  Changes

  - Iceberg.scala — added partitionColumnNames() to expose partition spec fields from Iceberg metadata
  - TableUtils.scala — insertPartitions() detects unpartitioned Iceberg tables and uses MERGE INTO instead
   of insertInto(). Removed saveMode parameter (always Overwrite).
  - BatchNodeRunner.scala — removed explicit saveMode argument (now handled internally)
  - ChrononMergeExtension.scala — new Spark session extension that adds WHEN NOT MATCHED BY SOURCE support
   to Spark 3.5's parser

  Tests (22 new in IcebergTest + 1 in IcebergMergeSyntaxTest)

  - Subrange overwrite: existing [01-08], write [03-05] → 03-05 replaced, rest untouched
  - Partial overlap extending beyond: existing [01-05], write [03-08] → overlap replaced, new days added
  - No overlap: existing [01-03], write [06-08] → both preserved
  - Full overlap: rewrite exact same range → all replaced, no duplicates
  - Shared file safety: job 1 writes [01-06], job 2 writes [04-07] → days 01-03 preserved in shared file
  - Stale row deletion: recompute produces fewer rows per day → old rows deleted, not preserved
  - DATE column type: partition overwrite works correctly with DATE values
  - TIMESTAMP column type: partition overwrite works correctly with TIMESTAMP values
  - INT column type: partition overwrite works correctly with INT values
  - ChrononMergeExtension syntax: ON FALSE + WHEN NOT MATCHED BY SOURCE parsed and executed correctly

  Checklist

  - Added Unit Tests
  - Covered by existing CI
  - Integration tested
  - Documentation update


nikhil-zlai (commented):
dumb question. why not delete incoming partitions and then write?

nikhil-zlai:
another question. should we actually partition unity managed tables?

tchow-zlai:
> another question. should we actually partition unity managed tables?

what do you mean?

tchow-zlai:
> dumb question. why not delete incoming partitions and then write?

it's hard to do that as a single transaction.

BenMagyar (approved):

nikhil-zlai (approved):

PR #841, opened 2025-06-06
remove legacy compile

tchow-zlai:
## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the compilation CLI into a single simplified `compile` command with minimal options.
  - Removed legacy compilation scripts and redundant internal logic to reduce complexity.
  - Updated CLI command registration to use the unified `compile` command instead of previous versions.

- **Chores**
  - Removed deprecated compilation files and updated test configurations to reflect the new compilation command.
  - Revised test suites to focus on direct function testing and metadata path correctness, replacing prior CLI invocation tests.

PR #277, opened 2025-01-27
cli v2

nikhil-zlai:
## Summary

Newer version with better ergonomics, and more maintainable codebase - smaller files, simpler logic - with little indirection etc with user facing simplification

- [x] always compile the whole repo
- [x] teams thrift and teams python
- [x] compile context as its own object
- [ ] integration test on sample
- [ ] progress bar for compile
- [ ] sync to remote w/ progress bar
- [ ] display workflow
- [ ] display workflow progress

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [x] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

- **New CLI Commands**
  - Introduced streamlined commands for synchronizing, backfilling, and deploying operations for easier management.

- **Enhanced Logging**
  - Improved colored, structured log outputs for clearer real-time monitoring and debugging.

- **Configuration & Validation Upgrades**
  - Strengthened configuration management and validation processes to ensure reliable operations.
  - Added a comprehensive validation framework for Chronon API thrift objects.

- **Build & Infrastructure Improvements**
  - Transitioned to a new container base and modernized testing/build systems for better performance and stability.

- **Team & Utility Enhancements**
  - Expanded team configuration options and refined utility processes to streamline overall workflows.
  - Introduced new data classes and methods for improved configuration and compilation context management.
  - Enhanced templating functionality for dynamic replacements based on object properties.
  - Improved handling of Git operations and enhanced error logging for better traceability.

PR #1765, opened 2026-04-21
refactor: Resolve output namespace at compile time; eliminate {{ db }} leaks

tchow-zlai:
## Summary

The upstream `{{ db }}` template stopped working in the zipline-ai fork when PR #841 (Jun 2025, "remove legacy compile") replaced `compilev2.py` with the current `cli/compile/parse_configs.py` pipeline and never wired through the old `set_templated_values` step. cli-v2 (#277) had carried `{{ db }}` for Join bootstrapParts / metaData dependencies; #841 collateral-dropped it; PR #1691 later swept the orphan `fill_templates.py`. For StagingQuery the story is different — `{{ db }}` inside StagingQuery fields was **never** wired here; the fork inherited an upstream state where `_set_templated_values` was gated on `cls == api.Join`.

Since then, `utils.output_table_name` has been emitting `{{ db }}.<name>` as a fallback when `outputNamespace` was null, and the placeholder leaked through compile into Airflow dep specs, the hub UI, and Spark table reads. This PR closes that cycle with a single invariant: **every compiled config has a resolved `outputNamespace`; nothing emits `{{ db }}` in user-facing output.**

### Compile pipeline (one pass, one invariant)

1. `_propagate_namespace` — pulls team default onto `metaData.outputNamespace` for the config and every nested config (JoinParts, JoinSource, ModelTransforms.models).
2. `_require_output_namespace` — hard check; fails compile if any node still has a null/empty `outputNamespace`.
3. `_resolve_namespace_placeholders` — walks the tree and substitutes an internal placeholder token in every string field that can carry it.

### Authoring-time `.table` access

`output_table_name(obj, full_name=True)` with null `outputNamespace` now emits an internal placeholder token (`_chronon_namespace_placeholder_`) rather than:
- Loading `teams.py` from disk (previous `_ensure_output_namespace_from_team` workaround — removed as leaky: filesystem I/O, console prints, cwd-dependent assertion, cycles on non-CLI entry points).
- Or raising (would break the common pattern of referencing another config's `.table` in an f-string when that config relies on team-default namespace inheritance).

The token is alphanumeric+underscore so it survives `utils.sanitize` unchanged — a `{{ db }}`-style token would be mangled to `____db____` inside `airflow_helpers.create_airflow_dependency`'s `name` field and escape substitution.

### Compile-time substitution sites

`_resolve_namespace_placeholders` resolves the token in each object's own post-propagation `outputNamespace`, across:
- `Source.events.table`
- `Source.entities.snapshotTable` / `mutationTable`
- `Join.bootstrapParts[].table`
- `StagingQuery.query` and `StagingQuery.setups`
- `StagingQuery.tableDependencies[].tableInfo.table`
- `metaData.customJson` (StagingQuery's Python wrapper bakes Airflow dep specs at construction time; this substitutes in the serialized JSON string)

Recurses into nested configs so each resolves against its own team's namespace.

### `.table` accessors

All three converge on `utils._ensure_name_and_get_output_table`:
- `GroupBy.table`
- `Join.table`
- `StagingQuery.table`
- `ModelTransforms.table`
- `join_part_output_table_name(join, jp, full_name=True)`

### Dead-code cleanup

- `join_part_output_table_name` had no production callers (one test + one sample fixture). The Scala side's `Extensions.partOutputTable` is the authoritative path.
- `_ensure_output_namespace_from_team` removed.
- Prior branch iterations' `applyNamespaceMacro` / `quote` flag on `ParametricMacro` reverted.

### Canary example that exercises the placeholder round-trip

- `staging_queries/gcp/team_default_ns_example.py` — StagingQuery with no explicit `output_namespace=` (relies on team default `data`).
- `joins/gcp/team_default_ns_join.py` — Join consuming `team_default_ns_example.v1.table` at authoring time.

Prior to this PR the `.table` reference on the first would have emitted `{{ db }}.<name>` and leaked into the Join's compiled Thrift. Now it emits the internal token, which `_resolve_namespace_placeholders` substitutes to `data.<name>` at compile.

### Test coverage

- Unit tests (`test_parse_teams.py`, `test_compile.py`) for the invariant, placeholder substitution paths, and token-shape variants.
- Integration test (`test/integration/test_canary_compile.py::test_canary_compile_produces_no_db_placeholder`) runs full `zipline compile` against canary and asserts no compiled thriftjson contains either the historical `{{ db }}`-shaped token or the internal placeholder.
- `CONFIG_INFOS` extracted as a module-level canonical list in `compile_context.py`; the test-only `user_authoring_folders()` filter lives inline in `test_compile.py`.

## Expected compiled diff when users re-run `zipline compile`

Small and strictly directional:
- **Explicit `output_namespace=` everywhere:** zero diff.
- **Relies on team defaults + `.table` cross-references:** `{{ db }}.x` → `<team_ns>.x` (fixing a broken value, not churn).
- **Missing `outputNamespace` with no team default:** compile error with a clear message pointing at `teams.py`. Strictly better than silent leakage.

## Checklist

- [x] Unit tests for propagation + strict invariant.
- [x] End-to-end compile tests for placeholder resolution.
- [x] Integration test: canary compile ships zero placeholder tokens.
- [x] 371 Python unit tests pass, 8 skipped (pre-existing).
- [x] Scala `api.compile`, `spark.compile`, `cloud_gcp.compile`, `cloud_azure.compile` clean.
- [x] Canary end-to-end compile clean.


BenMagyar:
I remember there being an issue where `outputNamespace` would not be inherited correctly off both: 

- `outputNamespace` being inherited from the default team to a new named team 
- `outputNamespace` not being inherited off of a team config 

in the case where we drop the `{{ db }}` name (which is already broken) and instead require this, do we hit some odd case with the method for getting output table names that is not using the lazy `.table` property?

tchow-zlai:
> I remember there being an issue where `outputNamespace` would not be inherited correctly off both:
> 
> * `outputNamespace` being inherited from the default team to a new named team
> * `outputNamespace` not being inherited off of a team config
> 
> in the case where we drop the `{{ db }}` name (which is already broken) and instead require this, do we hit some odd case with the method for getting output table names that is not using the lazy `.table` property?

I think I addressed this in the latest commit @BenMagyar

BenMagyar (approved):

nikhil-zlai (approved):

PR #1768, opened 2026-04-22
Allow injecting of Flink env vars for secrets injection + refactor status providers

piyush-zlai:
## Summary
Couple of Flink AKS related follow ups:
* AKS EventHub requires us to access Kafka using SASL where the sasl.jaas.config contains something like `org.apache.kafka.common.security.plain.PlainLoginModule required username="$ConnectionString" password="Endpoint=sb://zipline-demo-events.servicebus.windows.net/;SharedAccessKeyName=canary-flink-jobs;SharedAccessKey=foo=";` (where foo is a real secret). Rather than hardcoding in the teams.py, we allow users to specify the secret like: `FLINK_SASL_JAAS_CONFIG_VAULT_URI` and have it injected using the rails built in this PR as a pod env var. 
* Updated the Kafka src to read this env var if the jaas.config is empty (as there are other kafka cases where the jaas config doesn't contain secrets so safe for teams.py)
* Also brought in the Flink status provider class that we have in our internal platform repo to cut duplication + wired it to the submitter main methods.

## Testing
Confirmed that I am able to submit an AKS flink job which is looking up the jaas secrets via env (and these aren't logged in the jm / tms). 

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Flink health checks now use checkpoint-based validation with synchronous status resolution.
  * FLINK_* env vars are propagated into cloud pod deployments; SASL JAAS can be supplied via env or Vault (new FLINK_SASL_JAAS_CONFIG_VAULT_URI).
  * Kubernetes-backed Flink status provider added for REST-based job ID and checkpoint queries.

* **Tests**
  * New and expanded tests for Flink health logic and env-var injection/processing.

* **Chores**
  * Updated cloud submitters and build dependencies to support these behaviors.


tchow-zlai:
Reviewed the latest changes (`a5b813256`) — looks good overall. Checked out, compiled cloud_k8s/cloud_aws/cloud_azure/flink, and ran `cloud_k8s.test` (13/13) + `flink.test.testOnly KafkaFlinkSourceSpec` (11/11) all green.

### Strengths

- **`K8sFlinkStatusProvider`**: clean extraction, cloud-agnostic (uses ingress URL), `>= 3` completed-checkpoints heuristic is reasonable for startup health. Parsing is defensive (`Try { ... }.toOption.flatten`).
- **FLINK_-prefix envVar routing** in `K8sFlinkSubmitter.submit` (lines 299-308): well-scoped — only `FLINK_*` keys become pod env, prefix stripped so `FLINK_SASL_JAAS_CONFIG` surfaces as `SASL_JAAS_CONFIG` to the Kafka source. Explicit comment about keeping these out of `flinkConfiguration` (and hence out of Flink's startup-config logs) is the right call for secrets.
- **`KafkaFlinkSource.resolveSaslJaasConfig`**: guards on both SASL-not-in-protocol and already-set-jaas-config; handles whitespace-only existing values. Good test coverage.
- **`AzureSubmitter` / `EmrSubmitter` / `EmrServerlessSubmitter`** main methods now all wire in the shared `K8sFlinkStatusProvider` — nice de-duplication.

### Questions / minor

1. **`K8sFlinkStatusProvider.client` is a `lazy val` creating a Vertx instance that's never closed.** In the per-process `main` path this is fine (Await + JVM exit), but if the provider is ever constructed inside a longer-running Vertx-based process, the embedded `Vertx.vertx()` + `WebClient` pair will leak. Consider either (a) passing a `Vertx` in via constructor so the caller owns lifecycle, or (b) exposing a `close()` method. Not blocking for now since current callers are CLI `main` only.

2. **`RequestTimeoutMs = 10000L`** — is 10s tight enough for a slow/unreachable Flink JM? The `Await.result(..., 30.seconds)` at the call sites gives headroom, so this is fine; just flagging that if the request hangs past 10s the WebClient already times out.

3. **`isHealthyBasedOnCheckpoints >= 3`** threshold is a magic number. Worth a comment explaining why 3 (presumably: survive one checkpoint failure + first-completed + buffer?). Not critical.

4. **`AzureSubmitter.apply` at line 230-231** creates a new `ExecutionContext.fromExecutorService(Executors.newCachedThreadPool())` and never shuts it down. Same as (1): fine for CLI single-shot, bad shape if ever reused.

5. **Pod env strip of `FLINK_` prefix** assumes callers pass the already-resolved secret under the key `FLINK_SASL_JAAS_CONFIG` (not `FLINK_SASL_JAAS_CONFIG_VAULT_URI`). The teams.py sets `FLINK_SASL_JAAS_CONFIG_VAULT_URI` — so somewhere upstream (platform) has to translate `_VAULT_URI` → secret and rewrite the key to `FLINK_SASL_JAAS_CONFIG` before calling submit. Worth a scaladoc on `envVars` in `K8sFlinkSubmitter.submit` noting that contract, so a future caller doesn't ship the `_VAULT_URI` literal by accident.

Nothing blocking — LGTM with the above as follow-ups if you want.

david-zlai (approved):
if @tchow-zlai  had anymore followups

piyush-zlai:
@tchow-zlai - getting back on your qs for posterity:
* K8sFlinkStatusProvider.client is a lazy val - we currently use this as a singleton or in the cli invocation context so not as much of an issue.
* 10s request timeout - ideally we'd wrap that with a couple of retries so that we get a retry or two in before the 30s await timeout. 10s should be plenty for the Flink JM to respond so if we time out it is likely a network blip / jm not up. We do retry these checks at a higher level though - we keep checking for 10 or 15 mins before we fail so hopefully should be fine.
* isHealthyBasedOnCheckpoints >=3 - yeah bit of a magic number. Wanted to have a couple of successful checkpoints before we consider things good as in the past I have seen the occasional job where a checkpoint will pass and another will fail. So thinking was if we get a couple successful that should be a decent proxy for job being up.
* Pod env strip of FLINK_ prefix -> yea relying on the upstreams to adhere to the [Chronon Secrets Management](https://zipline.ai/docs/authoring_features/SecretsManagement) convention where they look up the secret and strip the vault_uri suffix.

tchow-zlai (approved):

PR #1782, opened 2026-04-29
feat: Add Flink Application Mode deployment for Dataproc

JoeTime17:
## Summary
  ### Background

  The current Flink deployment on Dataproc uses YARN Per-Job Mode via the Dataproc `FlinkJob` proto API, which internally runs `flink run`. This has several issues:

  1. **Per-Job Mode is deprecated since Flink 1.15**: Flink logs `Job Clusters are deprecated since Flink 1.15. Please use an Application Cluster/Application Mode instead.`
  2. **Master node resource pressure**: In per-job mode, the Flink client (driver JVM) runs on the Dataproc master node for the entire lifetime of the streaming job. For long-running streaming jobs, this means each driver JVM sits on the master forever, consuming resources. In production, this has caused memory and stability issues on the master node. With Application Mode, the master just submits and walks away — `FlinkJob.main()` runs inside the YARN JobManager container on a worker node instead
  3. **Dataproc FlinkJob API limitation**: The Dataproc `FlinkJob` proto has no field to control the deployment target. Setting `execution.target=yarn-application` in properties fails with `No ExecutorFactory found` because `flink run` doesn't load the `yarn-application` executor factory. The proto is hardwired to per-job mode with no way to change it

  ### Changes

  - Add support for Flink YARN Application Mode by introducing `FlinkApplicationLauncher`, which runs as a Dataproc `HadoopJob` to bypass the `FlinkJob` proto limitation. The launcher stages JARs from GCS to HDFS, executes `flink run-application -t yarn-application`, and monitors the YARN application lifecycle with a shutdown hook for graceful cancellation
  - Add `--flink-deployment-mode` CLI parameter (`default` | `application`) to select the deployment mode, defaulting to the existing per-job behavior for backward compatibility
  - Use `yarn.provided.lib.dirs` to manage additional JARs (cloud_gcp_lib_deploy.jar, Spark catalyst/SQL deps) on the YARN container classpath, with Flink distribution JARs staged to HDFS once and reused across deployments
  - Update `status()` and `emitJobRuntimeMetrics()` in `DataprocSubmitter` to use label-based Flink job detection (`job-type` label) instead of proto type checks, ensuring compatibility with both `FlinkJob` and `HadoopJob` proto types

  ## Checklist
  - [x] Added Unit Tests
  - [x] Covered by existing CI
  - [x] Integration tested
  - [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added a --flink-deployment-mode option (choices: "default", "application"); default behavior unchanged unless application mode is selected.
  * In application mode, Dataproc submits Flink as a YARN application via a launcher that stages artifacts, submits and monitors the YARN app, and handles staging/cleanup.

* **Tests**
  * Added tests covering application-mode job construction, launcher argument parsing, and command-building behavior.


cloud_gcp/src/main/scala/ai/chronon/integrations/cloud_gcp/DataprocSubmitter.scala
```diff
@@ -105,7 +105,7 @@ class DataprocSubmitter(jobControllerClient: JobControllerClient,
     try {
       val job: Job = jobControllerClient.getJob(projectId, region, jobId)
 
-      val isFlinkJob = job.getTypeJobCase == Job.TypeJobCase.FLINK_JOB
+      val isFlinkJob = job.getLabelsMap.asScala.get(JobType) == Some(FlinkJobType)
```
piyush-zlai: some of our users have jobs already submitted and running. Can we update this check to be a bit more defensive? If the get type is flink OR label is set to flink then consider the job to be a flink job.

piyush-zlai (commented):
A q - do we need to make any changes to the kill() method to handle the new hadoop job based flow?

Couple of other minor comments, largely lgtm.

JoeTime17:
> A q - do we need to make any changes to the kill() method to handle the new hadoop job based flow?
> 
> Couple of other minor comments, largely lgtm.

no, kill() also works on Hadoop Jobtype job. I have tested it by restarting a streaming with different `kv_concurrency` and log shows 
```
Cancelled running Flink job with id xxxxxxx for GroupBy name xxxxxxxxxxxx.
```
and the responding job is canceled and then the new job is running

piyush-zlai (approved):

PR #1784, opened 2026-04-29
fix(flink): defer extraFlinkConfig resolution to first submission

tchow-zlai:
## Summary

Customers running the hub on AKS without any streaming/Flink workflows were seeing the hub crash at startup with:

```
java.lang.IllegalArgumentException: FLINK_AZURE_CLIENT_ID must be set to
  the client ID of the zipline-flink-sa managed identity
    at ai.chronon.integrations.cloud_azure.AksFlinkSubmitter$.$anonfun$extraAzureFlinkConfig$1(AksFlinkSubmitter.scala:43)
    at ai.chronon.hub.AzureWorkflowExecutionVerticle.buildContext(...:124)
```

Root cause: `AksFlinkSubmitter.apply` eagerly evaluates `extraAzureFlinkConfig(env)` at construction time, and the `K8sFlinkSubmitter` constructor parameter is a strict `Map[String, String]`. So even when no Flink job ever runs, the hub fails to come up if `FLINK_AZURE_CLIENT_ID` / `FLINK_AZURE_TENANT_ID` aren't set.

## Fix

`K8sFlinkSubmitter.extraFlinkConfig` is now a **by-name** parameter, memoized internally via a private `lazy val resolvedExtraFlinkConfig`. Resolution happens on first call to `buildFlinkConfiguration` — i.e., the first actual Flink submission — rather than at hub startup.

```scala
class K8sFlinkSubmitter(
    flinkImage: String,
    buildInitContainerSpec: ...,
    extraFlinkConfig: => Map[String, String] = Map.empty,   // by-name
    ...
) {
  private lazy val resolvedExtraFlinkConfig: Map[String, String] = extraFlinkConfig
  ...
  // Used in buildFlinkConfiguration
  ) ++ flinkMemoryConfig(tier) ++ resolvedExtraFlinkConfig ++ jobProperties
}
```

Behavior change is strictly relaxing:
- Customers who don't run Flink never trip the env-var check.
- Customers who do run Flink see the **same** `IllegalArgumentException` with the **same** error message — just at submission time instead of hub startup.

## Test plan

- [x] All existing tests pass (`cloud_k8s.test` 13/13, `cloud_azure.test.testOnly AksFlinkSubmitterTest` 18/18).
- [x] New regression tests:
  - `AksFlinkSubmitter` constructs even when `FLINK_AZURE_*` env vars are missing.
  - `buildFlinkConfiguration` throws when those vars are unset.
- [ ] Customer can confirm hub starts on their AKS deployment without setting the env vars.

## Summary by CodeRabbit

* **Tests**
  * Added unit tests verifying lazy evaluation of Azure FLINK configuration, confirming startup succeeds when environment variables are absent and errors occur only upon configuration build.

* **Refactor**
  * Configuration values now evaluated lazily at first use rather than at startup, preventing initialization failures when required environment variables are unavailable.


piyush-zlai (approved):
Thanks for fixing!

PR #1786, opened 2026-04-29
Enable transforming key values in GroupBy aggregations

BenMagyar:
## Summary

The motivation behind this PR is that we'd like to usually apply some transform over a query key for aggregations, and then re-apply that transformation at inference time to fetch the feature value. We have two options (I think), either materialize the transform applied to each source in a StagingQuery, and then re-apply the transform externally from the caller of the fetcher service (obvious source of inconsistency) or push the transform onto the GroupBy itself and resolve where the transform should occur. 

Given an arbitrary transform of a key on a GroupBy though we must:

- Require that the GroupBy aggregation happens over that key transform
- Require that the streaming GroupBy aggregation happens over that key transform
- Require that the fetcher knows for a given GroupBy the key must be transformed prior to fetching the result
- Require that when running a Join, the raw key value must be transformed to match the existing pre-transformed value in the GroupBy. 

Off the primary motivation being how do you transform strings, we'd want NLP support which mostly comes from UDFs so we also need to include setups support within the GroupBy. 

The implementation of a transform would then look something like:

```py
query_item_views = GroupBy(
    sources=[search_queries],
    # Transform query to the lowercase query. Apply this transform offline and online
    keys={"query": "lower(query)"},
    aggregations=[
        Aggregation(
            input_column="item_id",
            operation=Operation.COUNT,
            windows=["7d"],
        )
    ]
)
```

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Group-by operations now support key transformations before aggregation
  * Join operations improved to handle key mapping with derived expressions

* **Bug Fixes**
  * Enhanced consistency in setup SQL execution across group-by and join workflows
  * Improved join request key derivation for complex join scenarios

* **Tests**
  * Added comprehensive test coverage for key transformation workflows
  * New test suite validating join operations with overlapping field mappings

### DIFF
--- a/api/src/main/scala/ai/chronon/api/Builders.scala
+++ b/api/src/main/scala/ai/chronon/api/Builders.scala
@@ -144,7 +144,9 @@ object Builders {
         keyColumns: Seq[String] = null,
         aggregations: Seq[Aggregation] = null,
         accuracy: Accuracy = null,
-        derivations: Seq[Derivation] = null
+        derivations: Seq[Derivation] = null,
+        keyTransforms: Map[String, String] = null,
+        setups: Seq[String] = null
     ): GroupBy = {
       val result = new GroupBy()
       result.setMetaData(metaData)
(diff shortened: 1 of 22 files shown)


BenMagyar:
Closing in favor of #1809

PR #1800, opened 2026-05-01
GroupBy planner: exclude onlineStrategy from batch node semantic hashes

nikhil-zlai:
## Why

While reviewing PR #1627 we noticed that `GroupByOps.semanticHash`
(`api/src/main/scala/ai/chronon/api/Extensions.scala:561`) unsets
`onlineStrategy` before hashing, with a comment noting it's online-only and
shouldn't trigger batch reprocessing.

That hash is correctly used by \`RelevantLeftForJoinPart\` and
\`JoinOps.semanticHash\` for join-level batch invalidation. **But the planner-side
node hash (computed in \`ConfPlanner.toNode\` from \`GroupByPlanner.semanticGroupBy\`)
does NOT unset \`onlineStrategy\`.** That hash drives \`BatchNodeRunner\`'s table
archival logic, so today flipping a GroupBy's \`onlineStrategy\` (e.g. between
DEFAULT and STREAMING_MEGATILES) flips every batch node's semantic hash —
forcing BatchNodeRunner to archive every batch table and re-run every batch job
downstream of the GroupBy, even though the underlying batch outputs are
byte-identical.

This is the inverse of what CodeRabbit flagged on #1627 — they thought the
\`Extensions.scala\` hash was masking online-state invalidation. Tracing actual
consumers shows: online-state invalidation goes through the planner-side node
hash (which includes \`onlineStrategy\`), so it works correctly. The bug is on
the *batch* side, where the planner hash *also* includes \`onlineStrategy\` and
shouldn't.

## What this PR changes

\`GroupByPlanner\` now uses two semantic hash variants:

- \`semanticGroupBy(...)\` (existing, unchanged) — for the streaming node, which IS
  structurally sensitive to online strategy.
- \`semanticGroupByForBatch(...)\` (new) — calls the existing helper and additionally
  unsets \`onlineStrategy\`. Used by backfill, upload, uploadToKV nodes.

## Test

New regression test \`GB planner batch nodes must keep the same semantic hash when
only onlineStrategy flips\`:

- Confirmed it fails on the unfixed code with \`"c[24516]" did not equal "c[367b7]"\`.
- Passes with the fix.
- Asserts streaming node hash STILL changes when \`onlineStrategy\` flips.

Full \`api.test\` suite: 229/229 green.

## Severity

Medium — only triggered when a user flips \`onlineStrategy\` on an existing
GroupBy (rare, typically a one-time migration). Cost is wasted batch compute,
not wrong data. Worth fixing to keep migrations cheap.


nikhil-zlai:
Folded into #1627 directly — commit `20f137a` on `nikhil/megatile`. Closing this standalone PR.

PR #1805, opened 2026-05-03
BaseFlinkJob: default runMegaTiledGroupByJob to throw, not abstract

nikhil-zlai:
## Why

While reviewing #1627 (MegaTile), reviewer feedback flagged
\`runMegaTiledGroupByJob\` as a high-risk **compile break** for custom subclasses
outside this repo. The method is declared abstract on \`BaseFlinkJob\`, even
though its scaladoc says *"Default delegates to buildMegaTiledTail"* — the doc
and the code disagree.

In-tree subclasses (\`FlinkGroupByStreamingJob\`, \`ChainedGroupByJob\`) implement
the method, so the chronon repo itself compiles. But any vendor / fork with a
custom \`BaseFlinkJob\` subclass would break compile against the new JAR until
they explicitly opted into MegaTile.

## Change

Replace the abstract declaration with a default implementation that throws
\`UnsupportedOperationException\` with an actionable message naming the subclass,
the GroupBy, and the offending config flag.

\`\`\`scala
def runMegaTiledGroupByJob(env: StreamExecutionEnvironment): DataStream[WriteResponse] =
  throw new UnsupportedOperationException(
    s"\${getClass.getSimpleName} does not implement runMegaTiledGroupByJob. " +
    s"GroupBy '\$groupByName' has onlineStrategy=STREAMING_MEGATILES, but the Flink job " +
    "subclass dispatched for it doesn't support mega-tiled execution. Override " +
    "runMegaTiledGroupByJob to opt in, or unset STREAMING_MEGATILES on this GroupBy."
  )
\`\`\`

## Why throw instead of silent fallback to \`runTiledGroupByJob\`

Chronon's serving correctness model: a misconfigured job should not silently
emit DEFAULT-mode output when the user explicitly asked for MegaTile. Loud-fail
at job startup is the right call.

## When this would actually trigger

| Scenario | Triggers? |
|---|---|
| User upgrades chronon JAR. No custom \`BaseFlinkJob\` subclass | No — in-tree subclasses already override |
| User has custom subclass; never enables MegaTile | No — method never called |
| User has custom subclass; enables \`STREAMING_MEGATILES\` on a GroupBy serviced by it | Yes — throws at job startup with a clear message |

## Tests

\`BaseFlinkJobMegaTileDefaultTest\` (new) verifies:
- A custom subclass that doesn't override gets \`UnsupportedOperationException\`
  whose message includes the subclass name, the GroupBy name, and \`STREAMING_MEGATILES\`.
- Subclasses that DO override are unaffected.

Full \`flink.test\` suite: 142/142 green. Aggregator: 70/70.


nikhil-zlai:
Folded into #1627 directly — commit `dc30f8f` on `nikhil/megatile`. Closing this standalone PR.

PR #1809, opened 2026-05-04
Use select in fetcher path to support transformed keys

BenMagyar:
## Summary

This PR is the alternative to #1786. Instead of pushing the transform up into the GroupBy and then resolving it as part of the Join + Fetcher, it extends the existing select usage on the Join to also work online in the fetcher. For a Join there are two considerations to make:

- Raw input value to mapped input value of a different name (`{ query_normalized: lower(query) }`) where the GroupBy is keyed by `query_normalized`. 
- Raw input value to mapped input value of the same name (`{ query: lower(query) }`) where the GroupBy is keyed to the same name. If this occurs, we will *recall the transformation online which is a breaking change* but behavior wise more correct. 

Any expression on the RHS would apply as a valid transform to be applied offline and online. This lets us have a single source of normalization (at least restricted to Chronon) over a key that is used in the GroupBy instead of having to additionally add the normalization to the online feature orchestration path pre-fetcher. 

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail fast with clearer errors when required join request keys are missing.
  * Input schema derivation now unions all configured input sources.

* **New Features**
  * Support for transformed/selected-key expressions for online joins with codec-aware key handling.
  * Join cache keys now include setup statements so different setups don’t share cached results.

* **Tests**
  * New integration and unit tests validating join key selection, codec behavior, and cache-key correctness.

* **Documentation**
  * Added “Transformed Join Keys” guidance and examples.


nikhil-zlai:
Performance-focused review:

1. Transformed key derivation now runs in `JoinPartFetcher.fetchJoins` before `GroupByFetcher` checks the batch IR cache. That puts expression parsing, serving-info lookup, `PooledCatalystUtil` wrapper creation, and first-use Catalyst plan construction on the request path, even for cache hits. This should be precomputed in the join metadata/codec cache so the hot path only executes an already-built transform when needed.

2. This only covers internal GroupBy join parts. External parts still call `part.applyMapping(joinRequest.keys)`, so selected left-key aliases will still be missing there unless callers send the derived key. Either apply the same translation to external parts or document/test that this PR is scoped to GroupBy parts.

nikhil-zlai (approved):
LGTM! thanks Ben!

tchow-zlai (approved):

PR #1824, opened 2026-05-06
Handle sparse batch node inputs and outputs 

nikhil-zlai:
## Summary
- remove the hard BatchNodeRunner input partition preflight and log missing input partitions as an orchestrator/sparse-output signal
- hard fail when required input tables are absent, so missing tables are not hidden as sparse partitions
- treat empty left/source-side ranges as sparse no-ops across BatchNodeRunner monolith joins, legacy `Driver.scala` join backfills, standalone `UnionJoin.computeJoinAndSave`, and modular join bootstrap/part/merge/derivation stages
- keep absent downstream input tables as real failures by scanning required tables directly; only existing tables with no rows in the requested range are skipped
- keep join stats compute skipping empty output ranges, but fail when the join output table itself is absent
- add regression coverage for required-table absence, missing stats output, sparse modular join no-ops, legacy join no-output handling, and UnionJoin sparse-left handling

## Verification
- `./mill __.reformat`
- `./mill 'spark[2.12.18].test.testOnly' ai.chronon.spark.batch.BatchNodeRunnerTest`
- `./mill 'spark[2.12.18].test.testOnly' ai.chronon.spark.join.UnionJoinTest`
- `./mill 'spark[2.12.18].test.testOnly' ai.chronon.spark.batch.ModularJoinTest -- -z "skip sparse downstream stages"`

## Summary by CodeRabbit

* **Bug Fixes**
  * Better recovery on job failure with safer rollback of archived outputs.
  * Jobs no longer fail on non-fatal post-job action errors.
  * Operations gracefully skip when expected input or join output is missing, avoiding exceptions.

* **Improvements**
  * More informative readiness and processing logs for missing input partitions and empty sources.
  * Additional test coverage for missing/empty-input scenarios to improve reliability.


spark/src/main/scala/ai/chronon/spark/batch/JoinBootstrapJob.scala
```diff
@@ -37,6 +37,10 @@ class JoinBootstrapJob(node: JoinBootstrapNode, metaData: MetaData, range: DateR
     // `f"${source.table}_${ThriftJsonCodec.md5Digest(sourceWithFilter)}"` Logic should  be computed by orchestrator
     // and passed to both jobs
     val leftDf = tableUtils.scanDf(query = null, table = leftSourceTable, range = Some(dateRange))
+    if (leftDf.isEmpty) {
```
kumar-zlai: @tchow-zlai as we were discussing offline ideally we shouldn't return, let it proceed to create empty output table but looks like we need a slightly more involved refactor to make these jobs return dataframe instead of write action. @nikhil-zlai is planning to address this as a follow up 

Also empty output is a valid case when the input is also empty but otherwise we should error out and good to make this change as well after the refactor

PR #1841, opened 2026-05-10
Only compute null rate for ts/ds cols

BenMagyar:
## Summary

Currently for timestamp columns during stats compute runs we're seeing:

```
Caused by: java.lang.IllegalArgumentException: Error while aggregating over 'created_date'. Date type and Timestamp time should not be aggregated over (They don't serialize well in avro either). Please use Query's Select expressions to transform them into Long.
```

As an alternative that was discussed we could skip computing stats for ts (and ds) columns as there is not much benefit to them. Here we restrict both column types to use `anyTransforms` type which restricts us just to null count for the column. 

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined metrics generation for date and timestamp columns to compute only essential null-related metrics.

* **Tests**
  * Added test cases validating enhanced statistics computation for datetime columns and numeric operations, with support for configurable time bucketing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zipline-ai/chronon/pull/1841)

### DIFF
--- a/aggregator/src/main/scala/ai/chronon/aggregator/row/StatsGenerator.scala
+++ b/aggregator/src/main/scala/ai/chronon/aggregator/row/StatsGenerator.scala
@@ -147,6 +147,8 @@ object StatsGenerator {
           dataType match {
             case _: api.ListType | _: api.MapType | _: api.StructType =>
               Seq.empty
+            case api.DateType | api.TimestampType =>
+              anyTransforms(name)
             case _ =>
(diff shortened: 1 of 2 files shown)


cristian-zlai (approved):
Thanks!

PR #1854, opened 2026-05-15
Add canary support for compilation and Zipline hub

david-zlai:
## Summary

Adds end-to-end canary support for the compile pipeline and the hub CLI:

1. **Thrift / API** — adds canary trio fields on `Team` (`canaryEnv`, `canaryConf`, `canaryClusterConf`), an `Environment` enum (`PROD`/`CANARY`), and a per-conf `environments` list on `GroupBy`/`Join`/`Model`/`StagingQuery`.
2. **Compile** — one `zipline compile` invocation now produces both `<root>/compiled/` (from `Team.env/conf/clusterConf`) and `<root>/canary_compiled/` (from `Team.canaryEnv/canaryConf/canaryClusterConf`). Strict mode isolation: prod compile never reads canary fields and vice versa. Decorative BEGIN/END banners + a mode-labeled "compilation successful" log make the two passes easy to distinguish.
3. **schedule-all** — gains `--env prod|canary` which filters confs by their `environments` field at scheduling time.
4. **Hub CLI** — `cancel`, `evaluate-schema`, and `list-tables` gain matching `--env` flags. The `DEFAULT_TEAM_METADATA_CONF` constant is replaced with `default_team_metadata_conf(env)` / `team_metadata_conf(team, env)` helpers so every env-aware command reads from the right output folder. A `print_env_banner` helper loudly states the target environment at the start of each env-aware command.

### Implementation notes

- **No skip logic** in canary compile — every team's confs land in `canary_compiled/`, even teams without their own canary fields. Those teams either inherit canary settings from the `default` team or end up with an empty `executionInfo` in their canary output.
- **`merge_team_execution_info`** picks the field tuple from a `_MODE_FIELD_NAMES` table keyed by `CompileMode`. That table is the single source of truth for field-set isolation between modes.
- **Two-pass orchestration** in `compile.py.__compile`: canary pass first, then prod. Each pass owns its own `CompileContext` / `Compiler`. JSON output stays prod-only to preserve the existing external contract; canary results are recoverable from disk.
- **`_get_airflow_deps_from_source`** now skips dep generation when no partition column is resolvable (the deep assert was firing in canary mode when a team had no `canaryConf`).

## Test plan

- [x] `./mill python.test` — full python suite (600+ tests, all green)
- [x] Added 6 tests in `python/test/test_canary_compile.py`:
  - canary folder is emitted
  - gcp Join with both trios lands in both folders with distinct `executionInfo`
  - canary team metadata uses canary fields
  - team without canary fields inherits only `default.canary*`
  - **no prod-only sentinel value leaks into `canary_compiled/`**
  - **no canary-only sentinel value leaks into `compiled/`**
- [x] Extended `python/test/canary/teams.py` (`gcp` Team) with the canary trio + `PROD_ONLY_SENTINEL_*` / `CANARY_ONLY_SENTINEL_*` markers used by the isolation tests
- [x] `./mill __.compile` — full repo compiles
- [x] Manual: ran `zipline compile` against `python/test/canary/`, confirmed both `compiled/` and `canary_compiled/` are produced with the expected per-mode `executionInfo`
- [ ] Manual: end-to-end check of `schedule-all --env canary` and the new `--env` flags on `cancel` / `evaluate-schema` / `list-tables` against a real hub

## Release Notes

* **New Features**
  * Added `environments` parameter to feature configurations (GroupBy, Join, Model, StagingQuery) to specify deployment targets (prod/canary).
  * Introduced canary compilation mode with separate output directory (`canary_compiled/`) for non-production deployments.
  * Added `--env` flag to CLI commands (schedule-all, cancel, eval-table, list-tables) for environment-specific operations.

* **Improvements**
  * Compilation now supports two-pass execution: canary first, then prod, enabling parallel environment management.
  * Team metadata selection is now environment-aware, using appropriate configs per deployment target.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zipline-ai/chronon/pull/1854)

### DIFF
--- a/python/src/ai/chronon/airflow_helpers.py
+++ b/python/src/ai/chronon/airflow_helpers.py
@@ -170,6 +170,16 @@ def _get_airflow_deps_from_source(source, partition_column=None):
         # Unknown source type
         return []
 
+    # Without a partition column we can't construct a meaningful dependency
+    # spec — skip rather than fail. This happens in canary compile when a team
+    # has no canaryConf (executionInfo.conf is empty) and the source's query
+    # also doesn't specify a partition column. In prod compile this same path
+    # is reached only when the user genuinely forgot to set the partition
+    # column anywhere, but failing later (at runtime / scheduling) is preferred
+    # over an obscure assertion deep in compile-time dep generation.
+    if source_partition_column is None:
+        return []
+
     return [
         create_airflow_dependency(table, source_partition_column, additional_partitions)
         for table in tables
--- a/python/src/ai/chronon/cli/compile/compile_context.py
+++ b/python/src/ai/chronon/cli/compile/compile_context.py
@@ -6,6 +6,7 @@ import ai.chronon.cli.compile.parse_teams as teams
 from ai.chronon.cli.compile.conf_validator import ConfValidator
 from ai.chronon.cli.compile.display.compile_status import CompileStatus
 from ai.chronon.cli.compile.display.compiled_obj import CompiledObj
+from ai.chronon.cli.compile.parse_teams import CompileMode
 from ai.chronon.cli.compile.serializer import file2thrift
 from ai.chronon.cli.formatter import Format
 from ai.chronon.cli.logger import get_logger, require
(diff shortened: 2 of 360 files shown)


david-zlai:
closing this for now in favor of 
https://github.com/zipline-ai/chronon/pull/1858

PR #1855, opened 2026-05-15
Add support for switching kubectl contexts for supporting different Zipline environments

david-zlai:
## Summary

Wires `--env prod|canary` into the admin CLI commands so they deterministically target the right Kubernetes cluster instead of relying on whatever `kubectl` context happens to be active. Prod and canary live in separate AWS accounts with the same `zipline-system` namespace, so without an explicit context flag it's easy to "think you're on canary" while actually running against prod.

### What's changed

- **Four admin commands gain `--env prod|canary` (default `prod`)**: `install`, `upgrade control-plane`, `doctor hub-health`, `doctor streaming-health`. Each prints a loud `🚀 PROD` / `🐤 CANARY` banner up front via the shared `print_env_banner` helper imported from `hub_runner.py`.
- **The two kubectl-driven commands (`control-plane`, `streaming-health`) also accept `--kube-context <name>`** and route every kubectl call through `--context <name>`. The context is resolved by `_kube_context_for_env`:
  1. `--kube-context` CLI override wins if passed
  2. Otherwise `ZIPLINE_PROD_KUBE_CONTEXT` / `ZIPLINE_CANARY_KUBE_CONTEXT` env var
  3. If neither is set, hard-fail with a setup hint (no silent fallback to the ambient context)
- **`admin.py` internals**: `_get_current_image` and `_upgrade_eks_services` now take a `context` arg; a new `_kubectl_argv(context, *args)` helper builds kubectl argv with `--context` baked in by construction so a missing context flag is impossible at every call site.
- **`admin_utils.py` internals**: `_run_kubectl` accepts an optional `context`; `run_infra_checks` and `_run_azure_infra_checks` use `functools.partial` to bind it once per function and forward via a local `kctl` alias, so the ~15 internal kubectl calls don't each need to repeat the kwarg.
- **Context resolution is deferred until after input validation** in `control_plane`, so a missing env var doesn't mask a more obvious arg error (e.g. wrong cloud, missing release).

### What didn't change

- Namespace: `zipline-system` stays the same for both envs (prod and canary deploy from the same Terraform — only the AWS account differs).
- `_EKS_SERVICES`, `admin_utils` hardcoded namespaces (`zipline-flink`, `flink-operator`).
- `upgrade data-plane` — already env-aware via `redeploy_streaming`'s path-based inference.
- AWS credential setup — still the user's responsibility (`aws eks update-kubeconfig --profile <prod|canary> --alias <ctx>`).

### Dependency

This PR depends on helpers added in #1854 (`print_env_banner` imported from `hub_runner.py`). It targets `david/add_canary_support` rather than `main`; rebase onto `main` once #1854 lands.

## Test plan

- [x] `./mill __.compile` — clean.
- [x] `./mill python.test python/test/test_admin.py` — 98/98 pass. Updates:
  - 16 `_upgrade_eks_services(...)` calls now pass `"test-ctx"`
  - 2 `upgrade control-plane` click tests pass `--kube-context test-ctx` and expect the new `(cloud, release, context)` signature
  - 1 image-tag positional-index test rewritten to scan kubectl argv by pattern rather than `cmd[4]` (since `--context X` shifts argv indices)
- [x] Manual smoke: `--env prod` and `--env canary` both hard-fail symmetrically when no context is configured, with the right env-var name in the error.
- [ ] Manual setup (one-time, post-merge):
  ```bash
  aws eks update-kubeconfig --name zipline-prod   --profile prod   --alias zipline-prod
  aws eks update-kubeconfig --name zipline-canary --profile canary --alias zipline-canary
  export ZIPLINE_PROD_KUBE_CONTEXT=zipline-prod
  export ZIPLINE_CANARY_KUBE_CONTEXT=zipline-canary
  ```
- [ ] Manual: `zipline admin upgrade control-plane aws --release 1.7.0 --env canary` — verify it hits the canary cluster, not prod.
- [ ] Manual: `zipline admin doctor streaming-health --env canary` — verify infra checks run against the canary cluster.


david-zlai (commented):
instead of doing this we can just prompt users for anything to be like:

zipline admin upgrade control-plane --env canary

"For the canary env, do you want to upgrade with this kubectl context _____ ?"


can do stuff in the future with --env flag. like set the env block in .zipline
```
canary
    kubeccontext: "canary123"
    artifact_prefix: "s3://...."
prod
    kubeccontext: "prod123"
    artifact_prefix: "s3://...."
```

david-zlai:
closing as i made the change already in https://github.com/zipline-ai/chronon/pull/1858

PR #1863, opened 2026-05-20
fix(spark): stabilize GroupByTest "group by from chaining gb"

bagipriyank:
## Summary
- Derive `today` from the last partition actually written by `DataFrameGen` instead of `System.currentTimeMillis()`.
- The previous form intermittently failed in CI when the test's wall-clock `today` did not match the generator's internal `today`, leaving the queried partition empty and producing a diff between the expected SQL result and the chaining `GroupBy` input.

## Checklist
- [ ] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Test plan
- `./mill spark[2.13.17].test.testOnly ai.chronon.spark.groupby.GroupByTest` — 3/3 green locally (without the fix: ~1/3 fails).
- `./mill spark[2.12.18].test.testOnly ai.chronon.spark.groupby.GroupByTest` — green locally.
- `./mill spark.checkFormat`.

PR #1886, opened 2026-05-29
fix(spark): stabilize HeterogeneousPartitionColumnsTest

bagipriyank:
## Summary

`HeterogeneousPartitionColumnsTest` flaked under parallel test load with:

```
java.lang.RuntimeException: No not-null keys found for key: item.
  at ai.chronon.spark.JoinUtils$.$anonfun$injectKeyFilter$5(JoinUtils.scala:309)
  at ai.chronon.spark.join.HeterogeneousPartitionColumnsTest.$anonfun$new$1(HeterogeneousPartitionColumnsTest.scala:157)
```

Reproduces under full parallel load (`./mill 'spark[2.12.18].test.testOnly' "ai.chronon.spark.join.*"`) but not in isolation — the classic timing-influenced flake signature.

Root cause turned out to be a stack of two issues working together:

1. **Sparse boundary partition.** `DataFrameGen.events` derives its time window from `CStream.TimeStream(minTs = currentTimeMillis() - window)`, so the first written partition only spans the fraction of a day from `minTs`'s time-of-day to midnight and is materially sparser than the rest. Combined with the 10% `StringStream` null rate, that boundary partition occasionally has zero non-null `item` keys.

2. **Engine-level partition-spec mismatch silently collapsed the compute range to that one sparse day.** Two latent bugs in the join engine made heterogeneous-partition joins quietly degrade rather than fail loudly:
   - `JoinUtils.getRangeToFill` built the left `PartitionRange` tagged with the left source's spec but populated `start` from `tableUtils.firstAvailablePartition`, which always normalizes results back to the `TableUtils` default spec. For heterogeneous joins this produced a mixed-format range (default-format start, custom-format end).
   - `TableUtils.unfilledRanges` compared `validPartitionRange.partitions` (in the input range's spec) against `outputExisting` (in the default spec). The set-diff across mismatched specs saw zero overlap and either left `missingPartitions` empty or collapsed the join's compute range to a single day.

The interaction is what made the test flaky: the engine consistently produced single-day computes, and `JoinUtils.injectKeyFilter` (in `runSmallMode`) collected keys from just that one day. When that day happened to be the sparse boundary with zero non-null `item` rows, it threw.

## Fix

Three changes, scoped narrowly:

1. **`JoinUtils.getRangeToFill`** translates `firstAvailablePartition` back into the left source's spec so the constructed `PartitionRange` has consistent start/end formats matching its tagged spec.
2. **`TableUtils.unfilledRanges`** canonicalizes `validPartitionRange` to the default spec before set operations, and drops a stray redundant translate inside the input-partitions loop (`TableUtils.partitions` already returns values in the default spec when `tablePartitionSpec` is supplied, so the previous code was double-translating).
3. **`HeterogeneousPartitionColumnsTest`** explicitly pins the left source's `startPartition` past the sparse boundary using the discovered written-partitions list (per @tchow-zlai's suggestion), fails fast via `getOrElse(throw ...)` if the test setup ever produces fewer than two partitions, and asserts the computed output spans multiple partitions — so a future regression in the engine's spec handling cannot silently collapse this test back to a one-day result.

The engine fixes are the root-cause fix and benefit any heterogeneous-partition consumer; the test-level pin is belt-and-suspenders that also documents the "skip sparse boundary" intent.

Same flake category as #1863 (stabilizing `GroupByTest "group by from chaining gb"`), but addressed at the engine level rather than via a per-test workaround so the heterogeneous-partition coverage the test was meant to exercise is preserved.

## Checklist

- [x] Added Unit Tests (covered by existing test, stabilized + multi-day range assertion added)
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Test plan

- `./mill 'spark[2.12.18].test.testOnly' "ai.chronon.spark.join.HeterogeneousPartitionColumnsTest"` — green in isolation (34s); now computes a 96-day range (`[20251115...20260219]`) instead of collapsing to one day.
- `./mill 'spark[2.12.18].test.testOnly' "ai.chronon.spark.join.*"` — 51/51 green under full 10-worker parallel load (~6m18s); previously failed deterministically on `HeterogeneousPartitionColumnsTest` in the same configuration.
- `./mill 'spark[2.12.18].test.testOnly' "ai.chronon.spark.{groupby,bootstrap,analyzer,batch}.*"` — 146/146 green (~11min), exercises `unfilledRanges` across the broader engine surface to confirm the canonicalization change doesn't regress default-spec callers.
- `./mill spark.checkFormat` — green.


tchow-zlai (approved):
LGTM thanks for the fix

PR #1860, opened 2026-05-20
[codex] Add GroupBy upload status endpoint

nikhil-zlai:
## Summary

- Add `GET /v1/groupby/:name/status` to return online GroupBy upload status.
- Surface `batchEndDate` from existing `GroupByServingInfo` through the Scala fetcher and Java wrapper.
- Add handler/fetcher tests and update fetcher API docs.

## Behavior

Successful responses return:

```json
{"groupByName":"team.group_by","batchEndDate":"2026-05-20"}
```

Offline GroupBys return a 400 with guidance to enable `online=True` and upload the GroupBy. Other retrieval failures return the existing `errors` shape with a 500.

## Validation

- `./mill -i online.test.testOnly ai.chronon.online.test.FetcherSchemaTest`
- `./mill -i service.test.testOnly ai.chronon.service.handlers.GroupByStatusHandlerTest`
- `./mill -i online.reformat`
- `./mill -i service.reformat`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * New API endpoint to retrieve online GroupBy upload status and batch end date; returns 200 on success, 400 for offline groupbys, 500 for other errors.

* **Documentation**
  * API docs updated with the new status endpoint, example requests, and response field descriptions.

* **Tests**
  * Added handler and fetcher tests covering success, offline (400) and generic failure (500) scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zipline-ai/chronon/pull/1860?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1891, opened 2026-06-01
Fix groupBy schema fetch to read from batch serving info

nikhil-zlai:
## Problem

`/v1/groupby/:name/schema` fails for valid online GroupBys (e.g. `gcp.user_activities.v1__1`) while `/v1/join/:name/schema` works for a join that has that GroupBy upstream.

Root cause: `Fetcher.fetchGroupBySchema` (added in #1835) reads the GroupBy **conf** from the metadata dataset (`CHRONON_METADATA`, key `group_bys/<name>`) purely to check `metaData.online` before reading the batch `GroupByServingInfo`. GroupBy confs are **not** uploaded to `CHRONON_METADATA`, so that read throws *"Couldn't fetch … group_bys/… Perhaps metadata upload wasn't successful"* and the endpoint errors out before it ever reaches the schema.

Join schema works because the Join conf **is** in `CHRONON_METADATA`, and it resolves each join part's GroupBy from the `<NAME>_BATCH` serving-info datasets (`buildJoinCodec`) — never needing a GroupBy conf in metadata.

## Fix

The batch `GroupByServingInfo` already embeds the full conf (`GroupByUpload.buildServingInfo` calls `setGroupBy`) and holds every schema the endpoint returns. So read the serving info directly from `<NAME>_BATCH` and derive `online` from the embedded conf, mirroring how join schema resolves GroupBy data. This removes the `CHRONON_METADATA` dependency from the GroupBy schema path.

Since `online` now comes from the TTL-cached serving info (the old `getConf` read was uncached), the offline branch triggers an async `getGroupByServingInfo.refresh(...)` so an offline→online flip converges within the refresh interval rather than the full TTL — matching the refresh-on-failure pattern in `buildJoinCodec`.

**Tradeoff:** a truly offline GroupBy with no batch upload now returns a generic "no serving info — ensure the batch upload succeeded" error instead of the friendly "enable online=True" message. The friendly message still shows when a serving info exists with `online=false`.

> Note: the `fetchGroupByStatus` endpoint added in #1860 still reads the conf from `CHRONON_METADATA` and has the same latent issue, but that's left out of scope for this PR.

## Testing

Extended the existing `GroupByUploadTest` decimal-types case, which already runs a **real** `GroupByUpload` via `OnlineUtils.serve` and bulk-loads the batch dataset **without** writing the conf to `CHRONON_METADATA` — i.e. the exact production setup. It now also asserts `fetchGroupBySchema` returns the expected key/value schema. Verified the assertion **fails against the old `getConf` path** with the same `Couldn't fetch group_bys/...` error and passes with the fix.

Updated the offline-groupBy unit test in `FetcherSchemaTest` to source `online=false` from the batch serving info (matching the new code path and the existing offline-status test there).

## Summary by CodeRabbit

* **Improvements**
  * Schema and online/offline status are now determined directly from serving info for more consistent schema resolution.

* **Bug Fixes**
  * Clearer failure when a GroupBy is offline; cache refresh is triggered to avoid stale reads.

* **Tests**
  * Updated tests to validate online/offline schema retrieval and assert resolved Avro key/value and derived fields.


nikhil-zlai:
Addressed the CodeRabbit note in bd7aef61: since online status now comes from the TTL-cached `GroupByServingInfo` (the old `getConf` read was uncached), the offline branch now triggers `getGroupByServingInfo.refresh(...)`. Note `TTLCache.refresh`/`force` update **asynchronously** and return the current cached value, so a same-call synchronous retry wouldn't see fresh data — instead this converges a subsequent request within the refresh interval rather than the full TTL, matching the refresh-on-failure pattern in `buildJoinCodec`.

piyush-zlai (approved):

sean-zlai:
@nikhil-zlai can you fix the conflict and merge?

PR #1913, opened 2026-06-08
JoinParts are not reused as no partitions are detected on clustered tables

BenMagyar:
## Summary

When a table is written to UC Iceberg, it is not partitioned so there are no partitions available when requested during the Join to determine which JoinPart tables must be recomputed (so every table gets recomputed). This gives us a fallback in the StepRunner only path to pull partitions as a query against the table. Given `ds` is often a string we can't effectively use file stats here to pull the partition info. 

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Improvements**
  * Added optional logical-partition derivation for Iceberg tables (disabled by default) to improve partition discovery for unpartitioned tables and improve unfilled-range calculations.
  * Updated join-part execution and partition-existence checks to enable logical-partition derivation and to use an explicit date range.
* **Tests**
  * Added coverage for unpartitioned Iceberg tables validating execution behavior and `unfilledRanges` results when logical partitions are enabled.

### DIFF
--- a/spark/src/main/scala/ai/chronon/spark/Driver.scala
+++ b/spark/src/main/scala/ai/chronon/spark/Driver.scala
@@ -1167,7 +1167,12 @@ object Driver {
           val partColumnName = spec.head._1
           val partitionSpec =
             PartitionSpec(partColumnName, tableUtils.partitionSpec.format, tableUtils.partitionSpec.spanMillis)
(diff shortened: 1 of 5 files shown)


BenMagyar:
This has been addressed in another PR that corrected readiness checks

PR #1915, opened 2026-06-09
fix(service): deduplicate OTLP resource attributes to prevent Microme…

RickyDepop:
fix(service): deduplicate OTLP resource attributes to prevent Micrometer crashing due to duplicate keys service.name

  Micrometer's OtlpConfig throws on duplicate resource attribute keys.
  Deduplicate via LinkedHashMap in buildOtlpResourceAttributes so
  OTEL_SERVICE_NAME can safely override the default service.name.

  Also read OTEL_SERVICE_NAME in OtelMetricsReporter.buildResource so
  one env var sets service.name for both the Micrometer (Vert.x) and
  OTel SDK (Chronon) metric pipelines.

## Checklist
- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Metrics exporter now deduplicates OTLP resource attributes, enforces predictable last-write-wins across configuration sources, and ensures a single service.name is emitted with explicit override from the service name environment variable. OpenTelemetry client behavior around resource attributes is more consistent and testable.

* **Tests**
  * Added Java and Scala tests validating attribute deduplication, precedence rules, and safe parsing/restoration of environment/system state.

### DIFF
--- a/online/src/main/scala/ai/chronon/online/metrics/OtelMetricsReporter.scala
+++ b/online/src/main/scala/ai/chronon/online/metrics/OtelMetricsReporter.scala
@@ -2,7 +2,7 @@ package ai.chronon.online.metrics
 
 import ai.chronon.online.metrics.Metrics.Context
 import io.opentelemetry.api.OpenTelemetry
-import io.opentelemetry.api.common.{AttributeKey, Attributes}
+import io.opentelemetry.api.common.{AttributeKey, Attributes, AttributesBuilder}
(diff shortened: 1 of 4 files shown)


piyush-zlai (approved):
Might be worth documenting in the fetcher README.md (docker/fetcher/README.md). Can merge when you think it's ready to go

piyush-zlai (approved):

PR #1929, opened 2026-06-11
fix(ci): use github context vars instead of git rev-parse in push-jars CTRL-236

kumar-zlai:
## Summary
#1926 refactored the per-jar \`aws s3 cp ... --metadata=\"...\$(git rev-parse HEAD)...\"\` into a top-level \`METADATA=\"...\$(git rev-parse HEAD)...\"\` assignment so we could iterate both buckets without repeating the metadata string. This job doesn't \`actions/checkout\` (only downloads artifacts), so git fails with \`not a git repository\`.

Bash semantics:
- Inline substitution **inside an argument** (\`--metadata=\"...\$(...)...\"\`): failure is silent, \`aws s3 cp\` runs with empty commit/branch values, step succeeds
- Substitution **inside a top-level assignment** (\`VAR=\"...\$(...)...\"\`): failure is the assignment's exit code, \`set -e\` aborts with exit 128

\`\`\`
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
Error: Process completed with exit code 128.
\`\`\`

## Fix
Use the \`github.sha\` + \`github.ref_name\` context vars — no git, no checkout needed, and the metadata is now *actually* populated (the pre-#1926 version was already uploading empty commit/branch values; the bug was just hidden by inline-substitution semantics).

## Validation
Reproduced the failure locally in a non-git dir:

\`\`\`
$ bash -c 'set -eo pipefail; METADATA=\"updated_date=\$(date),commit=\$(git rev-parse HEAD),branch=\$(git rev-parse --abbrev-ref HEAD)\"; echo \$METADATA'
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
exit=128                                            # ← reproduces CI

$ bash -c 'set -eo pipefail; GITHUB_SHA=abc GITHUB_REF_NAME=main; METADATA=\"updated_date=\$(date),commit=\${GITHUB_SHA},branch=\${GITHUB_REF_NAME}\"; echo \$METADATA'
updated_date=Thu Jun 11 21:14:01 UTC 2026,commit=abc,branch=main
exit=0                                              # ← Option B clean
\`\`\`

## Tracks
[CTRL-236](https://linear.app/zipline-ai/issue/CTRL-236)

## Summary by CodeRabbit

* **Chores**
  * Deployment workflow updated to derive artifact metadata from CI-provided commit SHA and branch name for more consistent and reliable tagging.
  * Upload metadata for canary and latest artifacts now uses CI-supplied commit/branch and current timestamp for S3 and GCS uploads, improving accuracy of release metadata and traceability.


chewy-zlai (approved):

cristian-zlai (commented):
There's a few other calls in the same workflow file... should we change all ?

kumar-zlai:
Good catch — addressed the gcloud `--custom-metadata` calls (L604–636), migrated to `${{ github.sha }}` / `${{ github.ref_name }}` same as the AWS fix. Left the docker tag at L336 on `$(git rev-parse --short HEAD)` since that job already does `actions/checkout` and the short form reads cleaner inline.

PR #1932, opened 2026-06-11
Drop online-flag gate from groupBy schema fetch

nikhil-zlai:
## Problem

After #1891 merged, `/v1/groupby/{name}/schema` still fails for GroupBys that back an online join (reported by Sean). The endpoint reaches the fixed code (`GroupBySchemaHandler` → `JavaFetcher.fetchGroupBySchema` → `Fetcher.fetchGroupBySchema`), reads the batch `GroupByServingInfo` correctly, and then throws:

```
IllegalArgumentException: GroupBy <name> is not online. ... enable online=True and upload the GroupBy.
```

Root cause: `fetchGroupBySchema` gates on `servingInfo.groupBy.metaData.online`. That flag is the **wrong signal**. A GroupBy that exists only as a join dependency is uploaded to `<NAME>_BATCH` (so the join can fetch it) yet commonly carries `online=false` in its own conf. Data fetching (`GroupByFetcher`), join part fetching (`JoinPartFetcher`), and `buildJoinCodec` all read the **same** `getGroupByServingInfo` and **none** check the online flag — so the GroupBy is fully servable, but the schema endpoint alone rejects it with a 400.

This was masked in #1891 because its test forced `online = true`.

## Fix

Drop the online gate. Presence of the batch serving info is the real signal: read it and return the schema. If it's absent, `getGroupByServingInfo` already surfaces a clear "make sure a batch upload was successful" error. This makes the schema endpoint consistent with data/join fetching.

## Testing

- `GroupByUploadTest` decimal-types case (a **real** `GroupByUpload` via `OnlineUtils.serve`, no conf in `CHRONON_METADATA`): flipped from `online = true` → `online = false`, so it now mirrors a join-dependency GroupBy. The data fetch and the schema fetch both resolve. Verified this case **fails before the fix** with the exact `GroupBy ... is not online` error.
- `FetcherSchemaTest`: the offline-conf case now asserts the schema **resolves** (regression for Sean's bug); added a case asserting a GroupBy with **no batch upload** fails. The unrelated `fetchGroupByStatus` tests from #1860 are untouched.

> Note: `fetchGroupByStatus` (#1860) still gates on `online` via a `CHRONON_METADATA` `getConf` read and has the same latent issues — left out of scope here.

## Release Notes

* **Bug Fixes**
  * Improved schema resolution for group-bys to reliably handle cases where batch serving information is available.

* **Tests**
  * Enhanced test coverage validating schema resolution succeeds with batch serving information present.
  * Added validation ensuring schema fetch fails appropriately when required serving information is unavailable.


cristian-zlai (commented):
It feels we are replacing one type of failure for another.
afaik we only orchestrate uploads i.e. group by serving info, for online group bys, and we enforce online joins to contain online group bys. 
I understand having a "check that an upload succeed" error (rather than a success as it has now) even if it's online, but isn't it true that a not online group by won't succeed?

nikhil-zlai:
> afaik we only orchestrate uploads i.e. group by serving info, for online group bys,
> but isn't it true that a not online group by won't succeed?

apparently not! at some point we started making it so that setting join online=True meant the groupBy will implicitly be online.

cristian-zlai (approved):

PR #1945, opened 2026-06-17
Fix modular join source readiness

nikhil-zlai:
## Summary

Modular joins could wait on a Chronon-generated intermediate table instead of the real source table.

Simple example:

```text
raw.events -> my_join__source -> my_join__join_part
```

Before, Chronon could create a sensor for `my_join__source`. But that table is produced by Chronon, so the workflow could get stuck waiting on the wrong thing.

After this PR, standard modular joins create sensors for the real external inputs, like `raw.events`, and skip intermediate tables produced inside the same modular plan.

This also fixes `SourceJob` date handling for source tables that use a custom partition column, such as `event_date` instead of `ds`.

## Tests

- `JAVA_HOME=/tmp/jdks/jdk-17.0.19+10 PATH=/tmp/jdks/jdk-17.0.19+10/bin:$PATH ./mill --no-daemon -j 1 api.checkFormat + api.test.testOnly ai.chronon.api.test.planner.JoinPlannerTest`
- `JAVA_HOME=/tmp/jdks/jdk-17.0.19+10 PATH=/tmp/jdks/jdk-17.0.19+10/bin:$PATH ./mill --no-daemon -j 1 spark.test.testOnly ai.chronon.spark.batch.SourceJobTest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced external source sensor identification in modular join planning to comprehensively evaluate all offline input tables, improving data dependency handling
  * Fixed source job execution to correctly apply source-specific partitioning during data scanning and iteration

* **Tests**
  * Added verification that join planner generates appropriate external source sensors for standard modular join configurations
  * Added validation of source job behavior with custom partitioning specifications

### DIFF
--- a/api/src/main/scala/ai/chronon/api/planner/JoinPlanner.scala
+++ b/api/src/main/scala/ai/chronon/api/planner/JoinPlanner.scala
@@ -242,6 +242,33 @@ class JoinPlanner(join: Join)(implicit outputPartitionSpec: PartitionSpec)
       statsComputeNodeOpt
   }
 
+  private def tableDependencies(node: Node): Seq[TableDependency] =
+    Option(node.metaData)
+      .flatMap(metaData => Option(metaData.executionInfo))
+      .flatMap(executionInfo => Option(executionInfo.tableDependencies))
+      .map(_.toScala.toSeq)
+      .getOrElse(Seq.empty)
+
+  private def externalSensorNodes(nodes: Seq[Node]): Seq[Node] = {
+    val internalOutputTables = nodes.flatMap(node => Option(node.metaData).map(_.outputTable)).toSet
+    val externalDeps = nodes
+      .flatMap(tableDependencies)
+      .filter { tableDependency =>
+        Option(tableDependency.tableInfo)
+          .flatMap(tableInfo => Option(tableInfo.table))
+          .exists(table => !internalOutputTables.contains(table))
+      }
+      .distinct
+
+    val sensorMetaData =
+      MetaDataUtils.layer(join.metaData, "sensor", join.metaData.name + "__external_sensors", externalDeps)
+
+    ExternalSourceSensorUtil
+      .sensorNodes(sensorMetaData)
+      .map(es =>
+        toNode(es.metaData, _.setExternalSourceSensor(es), ExternalSourceSensorUtil.semanticExternalSourceSensor(es)))
+  }
+
   def metadataUploadNode: Node = {
     val stepDays = 1 // Default step days for metadata upload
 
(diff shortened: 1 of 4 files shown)


BenMagyar (approved):

tchow-zlai (changes requested):

tchow-zlai (commented):
## De-dup of sensor nodes is by `TableDependency`, not by table

The new `externalSensorNodes` de-dups with `.distinct`:

```scala
val externalDeps = nodes
  .flatMap(tableDependencies)
  .filter { ... !internalOutputTables.contains(table) }
  .distinct
```

`TableDependency` is a Thrift struct whose `equals`/`hashCode` cover **all** fields, not just the table (`common.thrift`):

```
struct TableDependency {
  1: optional TableInfo tableInfo
  2: optional Window startOffset
  3: optional Window endOffset
  4: optional string startCutOff
  5: optional string endCutOff
  7: optional string semanticHash
  ...
}
```

So `.distinct` only collapses byte-for-byte identical deps. Two dependencies on the **same table with different ranges** (different `startOffset`/`endOffset`/cutoffs) are unequal objects and both survive. This is realistic in the standard modular path — e.g. two GroupBys reading the same raw table with different aggregation windows, or a raw table used both as the left source and inside a GroupBy.

Downstream, `ExternalSourceSensorUtil.sensorNodes` emits **one node per surviving dep**, named purely by table:

```scala
f"${td.tableInfo.table}__sensor"                  // node name
outputTableOverride = Option(td.tableInfo.table)  // output table
```

So two same-table/different-range deps yield **two sensor nodes with identical names and identical output tables** — a collision in a plan that keys nodes by name (`terminalNodeNames`), not just harmless redundancy. It also contradicts the agreed design: a sensor just detects the max timestamp of a table and assumes earlier partitions exist, so there should be exactly one sensor per table.

### Suggested fix — de-dup by table, squash to max end range

```scala
private def externalSensorNodes(nodes: Seq[Node]): Seq[Node] = {
  val internalOutputTables = nodes.flatMap(node => Option(node.metaData).map(_.outputTable)).toSet

  val externalDeps = nodes
    .flatMap(tableDependencies)
    .filter { td =>
      Option(td.tableInfo)
        .flatMap(ti => Option(ti.table))
        .exists(table => !internalOutputTables.contains(table))
    }
    // One sensor per table. Sensors only detect the latest partition, so the start
    // offset/cutoffs are irrelevant when collapsing — keep the max end range.
    .groupBy(_.tableInfo.table)
    .values
    .map(deps => deps.maxBy(endRangeOf))
    .toSeq

  val sensorMetaData =
    MetaDataUtils.layer(join.metaData, "sensor", join.metaData.name + "__external_sensors", externalDeps)

  ExternalSourceSensorUtil
    .sensorNodes(sensorMetaData)
    .map(es =>
      toNode(es.metaData, _.setExternalSourceSensor(es), ExternalSourceSensorUtil.semanticExternalSourceSensor(es)))
}
```

(`endRangeOf` = whatever ordering represents the max end across deps for a table.)

### The new test can't catch this

`JoinPlannerTest` uses four **distinct** tables (`left_events`, `dim_snapshot`, `dim_mutations`, `other_events`), so it never exercises the same-table/multiple-ranges path. And the assertion collapses into a `Set` of output-table strings:

```scala
val sensorOutputTables = plan.nodes.asScala.filter(...).map(...outputTableInfo.table).toSet
sensorOutputTables should equal(Set("test.left_events", ...))
```

A `Set` hides duplicate sensor nodes even if emitted. Please add a case where one table is depended on with two different ranges and assert on a **list/count** (exactly one sensor node per table) plus that the surviving dep carries the max end range.

### Minor

- `internalOutputTables` compares `node.metaData.outputTable` against `tableDependency.tableInfo.table`. This relies on both being fully-qualified (`namespace.table`) identically — worth confirming, since a qualified/unqualified mismatch would let an intermediate table leak through as an "external" sensor (the original bug).

The `SourceJob` custom-partition-column fix looks correct and is well covered by `SourceJobTest`. 👍

PR #1957, opened 2026-06-22
Validate end-ds is on/after start-ds in hub CLI

david-zlai:
## Summary
- Adds a pre-submission check in `hub_runner.py` that fails fast (with a clear message) when a user passes `--end-ds` earlier than `--start-ds`.
- Wired into the shared `end_ds_option` decorator so `backfill`, `run-adhoc`, and `clear-downstream` all inherit the check.
- Equal start/end is still allowed (single-partition backfills are legitimate).

## Test plan
- [x] New tests in `test_hub_runner.py` cover both the rejection path and the equal-date path.
- [ ] Manually verify `zipline hub backfill --start-ds 2024-02-15 --end-ds 2024-01-15 ...` aborts before posting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent backfill operations with an end date earlier than the start date. Invalid date ranges now abort with a clear error message instead of proceeding.

* **Tests**
  * Added test cases validating date range validation behavior for the backfill command.


nikhil-zlai (approved):

tchow-zlai (commented):
can you double check this doesn't break `run-adhoc` since those commands don't have `start-ds` ?

tchow-zlai (approved):

PR #1966, opened 2026-06-25
Handle daily catalog partitions returned as timestamps

nikhil-zlai:
## Summary
- Normalize catalog partition labels like `yyyy-MM-dd HH:mm:ss` to daily `yyyy-MM-dd` at catalog-input boundaries
- Keep strict `PartitionSpec.epochMillis` behavior for user-authored partition strings
- Cover unescaped GCP-style labels and Hive-escaped `00%3A00%3A00` labels in tests

## Testing
- `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 PATH=/usr/lib/jvm/java-17-openjdk-amd64/bin:$PATH ./mill --no-server api.test.testOnly "ai.chronon.api.test.PartitionSpecTest"`
- `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 PATH=/usr/lib/jvm/java-17-openjdk-amd64/bin:$PATH ./mill --no-server spark.test.testOnly "ai.chronon.spark.catalog.FormatTest"`
- `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 PATH=/usr/lib/jvm/java-17-openjdk-amd64/bin:$PATH ./mill --no-server spark.test.testOnly "ai.chronon.spark.other.TableUtilsTest" -- -z "normalize timestamp-shaped daily catalog partitions"`
- `JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 PATH=/usr/lib/jvm/java-17-openjdk-amd64/bin:$PATH ./mill --no-server __.reformat`

## Summary by CodeRabbit

* **New Features**
  * Added catalog partition parsing to normalize externally provided partition strings, including daily midnight timestamp formats and percent-encoded variants.
* **Bug Fixes**
  * Improved same-grid partition normalization and downstream watermark/end-time calculations by using canonicalized parsed partitions.
  * Updated “unparsed partitions” warning detection to rely on successful catalog-partition parsing.
* **Tests**
  * Added coverage for midnight acceptance/rejection, URL-encoded normalization, and timestamp-shaped daily catalog partitions.
  * Added end-to-end canary compilation test verifying `VERSION` is stamped from the process environment (defaulting to `latest`).


tchow-zlai:
Thanks for putting this together. I think the boundary placement is right: keep `epochMillis` strict for Chronon-authored partition strings, and normalize only values read back from catalogs/watermarks.

The part I would tighten is that `parseCatalogPartition` currently only accepts timestamp-shaped values when `format == PartitionSpec.daily.format`. That makes the fix depend on the default `yyyy-MM-dd` rendering even though the issue is catalog input normalization for any daily grid. I would also make the timestamp formatter strict, because `DateTimeFormatterBuilder` defaults to `SMART` and can coerce invalid dates.

A shape I think would be more robust:

```scala
import java.time.format.ResolverStyle

private val DailyCatalogTimestampFormatter: DateTimeFormatter =
  new DateTimeFormatterBuilder()
    .appendPattern("uuuu-MM-dd HH:mm:ss")
    .optionalStart()
    .appendFraction(ChronoField.NANO_OF_SECOND, 1, 9, true)
    .optionalEnd()
    .toFormatter(Locale.US)
    .withResolverStyle(ResolverStyle.STRICT)

def parseCatalogPartition(value: String): Option[String] = {
  if (value == null) return None

  val decoded =
    Try(URLDecoder.decode(value.replace("+", "%2B"), StandardCharsets.UTF_8.name()))
      .getOrElse(value)

  Seq(value, decoded).distinct.iterator
    .flatMap { candidate =>
      canonical(candidate).orElse {
        if (grid.isDaily) {
          Try(LocalDateTime.parse(candidate, DailyCatalogTimestampFormatter)).toOption
            .filter(_.toLocalTime == LocalTime.MIDNIGHT)
            .map(_.toInstant(ZoneOffset.UTC).toEpochMilli)
            .filter(ms => grid.leftBound(ms) == ms)
            .map(at)
        } else None
      }
    }
    .toSeq
    .headOption
}
```

Then I would add tests for:

- default daily: `2024-03-02 00:00:00` -> `2024-03-02`
- encoded daily: `2024-03-02 00%3A00%3A00` -> `2024-03-02`
- non-default daily: `PartitionSpec("ds", "yyyyMMdd", WindowUtils.Day.millis)` normalizes `2024-03-02 00:00:00` -> `20240302`
- non-midnight timestamp is rejected
- invalid dates like `2024-02-30 00:00:00` are rejected

That keeps the API contract strict while making the catalog shim format-independent within the daily-grid case.

tchow-zlai (approved):

PR #1991, opened 2026-07-02
Add keyFilter to GroupBy to shrink batch uploads

nikhil-zlai:
## Summary

Uploads can dominate serving costs for groupBys whose sources contain far more keys than are ever fetched. This adds `GroupBy.keyFilter` — an `EntitySource` whose snapshot partition for the upload date defines the active keyset. The aggregated upload output (one row per key) is semi-joined against the filter's distinct key tuples, so filtered-out keys never reach the KV store.

## Design decisions

- **Post-aggregation join**: whole-key filtering commutes with per-key aggregation, so filtering after aggregation is semantically identical — and it avoids semi-joining the raw (potentially windowed) input against a filter too large to broadcast. The temporal path's aggregation output is kryo-encoded `(KeyWithHash, FinalBatchIr)`, so it filters via an RDD join on the subset-key tuple, casting filter columns to the key schema's types (RDD joins don't coerce).
- **Upload-only, structurally**: the filter lives entirely in `GroupByUpload`; `GroupBy.from` is untouched. Backfills and the join → joinPart → groupBy path ignore it (there the join's left side already restricts keys).
- **Non-semantic**: `keyFilter` is stripped in `semanticGroupBy` (like topics), so setting/changing it never invalidates offline nodes or re-triggers the deploy chain. Unset filters leave existing semantic hashes byte-identical (TSimpleJSON omits unset optionals) — covered by a planner test.
- **Fail-closed**: a missing/empty filter partition, or a filter sharing no columns with `keyColumns`, fails the job rather than writing an empty upload (which would wipe batch data in the KV store).
- **Subset keys allowed**: the filter may provide any subset of `keyColumns` (e.g. only `user_id` for a `(user_id, merchant_id)` groupBy); provided columns are matched as a tuple.
- **Filter scanned at `endDs`** even for temporal accuracy — the shifted (`endDs+1`) partition doesn't exist yet when the upload for `endDs` runs. `partitionLag` on the filter's query can shift it if needed.

## Caveat

For TEMPORAL groupBys with a streaming topic, the streaming job still writes all keys, so filtered-out keys may serve partial streaming-only aggregates instead of nulls. Documented on the thrift field and python docstring.

## Testing

- `GroupByPlannerTest`: semantic hashes identical for all four nodes with the filter set vs unset.
- `GroupByUploadTest`: snapshot-accuracy test asserting the filtered upload contains exactly the filter's keys with byte-identical aggregates for survivors; temporal-accuracy test covering the RDD-join path the same way; failure tests for empty filter partition and no key-column overlap.
- Python `test_group_by.py`: `key_filter` construction (EntitySource or Source-wrapped), select/key intersection validation, events-source rejection.
- Full `GroupByUploadTest` (16/16), `GroupByTest` (13/13), `api.test`, python tests, and ruff all green.

## Summary by CodeRabbit

* **New Features**
  * Added optional `keyFilter` (entity source) to GroupBy to restrict batch upload outputs to keys from the selected snapshot partition (whole-key semi-join).
  * Snapshot and temporal uploads now support key-filtered writing.

* **Bug Fixes**
  * `keyFilter` is excluded from semantic hashing for GroupBy/Join and embedded join trees, preventing unnecessary retriggers.
  * Updated planner behavior and dependency wiring to distinguish upload vs backfill when `keyFilter` is present.

* **Validation / Safety**
  * Fail-fast checks for non-entity filters, mismatched key columns, and empty filtered key sets.
  * Added unit tests covering semantic-hash invariance and key-filter behavior.


piyush-zlai (commented):
Do we need to update the planner to add a dep on the new entity key filter dataset?

tchow-zlai (approved):

PR #2029, opened 2026-07-15
fix(JoinRequestKeys): fix online join key derivation for expressions that use fields from a struct

george-bennett-depop:
## Summary
Fixes online join key derivation for chaining when a left-side select expression
un-nests a value from a struct (e.g. `test_id = CAST(data.testId AS BIGINT)`).

Previously the fetcher collapsed the expression's inputs to the base column 
and typed it with the mapped GroupBy key's type and the struct was treated as the
key's type (i.e. `BIGINT` instead of `STRUCT`) and field extraction would fail. 

```Can't extract a value from "data". Need a complex type [STRUCT, ARRAY, MAP] but got "BIGINT"```

This should now track full extraction paths and synthesize a minimal
struct type from the fields the expressions actually access.

Worth noting, leaf types within this synthesized struct are still inferred from the mapped
GroupBy key's type. To compensate, type-mismatch failures
now raise a more descriptive error naming the join, the expression, and the
expected v actual type per request key.

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added support for deriving join request keys from nested input fields and struct-based select expressions.
  * Improved construction of complex request-key schemas from nested attribute paths.

* **Bug Fixes**
  * Enhanced validation with clearer, join-specific error messages when request key values have incompatible shapes or types.
  * Error details now include the relevant expressions and expected vs. actual JVM types.

### DIFF
--- a/online/src/main/scala/ai/chronon/online/fetcher/JoinRequestKeys.scala
+++ b/online/src/main/scala/ai/chronon/online/fetcher/JoinRequestKeys.scala
@@ -65,9 +65,32 @@ private[online] object JoinRequestKeys {
         if (keysToDerive.isEmpty) {
           Map.empty[String, Any]
         } else {
-          val derivedValues = catalystUtil
-            .map(_.performSql(request.keys).headOption.getOrElse(Map.empty))
-            .getOrElse(Map.empty)
+          // Input types are inferred, so mismatched caller values fail deep in Catalyst encoding -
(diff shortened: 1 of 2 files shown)


george-bennett-depop:
Closing as super seeded by #2030

PR #2032, opened 2026-07-15
Fix join-key projection in chained calls in flink path

BenMagyar:
## Summary

This issue happens because `fetchJoin` was re-applying the key projection to an already projected key. In the case of:

```
raw event: data.productId = "123"
left.query: product_id = CAST(data.productId AS BIGINT)
projected request to fetcher: product_id = 123L
GroupBy key fetch: product_id = 123L
 ```
 
we would re-apply the left.query projection again on the projected request that now only contains `product_id`. To solve for this we don't re-project and we be explicit about when we have vs have not projected. 

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added support for registering and using Hive UDFs in Flink group-bys and joins.
  * Added support for loading UDF JARs from S3 and GCS.
  * Added configurable AWS credential provider support for remote JAR loading.
  * Join derivations now support setup statements, enabling custom functions in derived keys and fields.
  * Added new canary configurations covering Hive UDF-based features and joins.

* **Bug Fixes**
  * Improved chained-join schema handling and deferred key-derivation failures until needed.

* **Documentation**
  * Documented the AWS credential provider configuration for remote JAR usage.


BenMagyar:
Closing we had gone a different path

PR #2037, opened 2026-07-16
Skip redundant GigaTile large recomputes on evictions between tail-hop boundaries

nikhil-zlai:
## Summary

Eviction ran `recomputeRunningLargeIr` unconditionally on every callback — decoding every retained daily slot and re-merging per column. Eviction timers fire on the small-hop cadence (minutes) while large tail hops span hours or a day, so for a 7-day window with hourly hops ~11 of 12 recomputes rebuilt a byte-identical IR. Keys awaiting a deferred publication pay the same full recompute per incoming event via `rebuildPendingAtCurrentTime`.

## Approach

Gate **only the recompute** — the emit/compare path is unchanged, so aggregator- and Flink-level emission semantics are identical:

- Reuse the boundary predicate the event path already trusts (`largeIrMayChangeSinceLastRecompute`): between tail-hop boundaries the sawtooth view is invariant, so a recompute rebuilds an identical IR.
- Add a per-key `largeIrDirty` flag for the one divergence the predicate cannot see: a replayed event the batch already covers double-counts in the incremental view until a recompute prunes its daily slot ("transiently double-count … corrected by eviction" contract). Set on batch-overlap events (`eventTs < batchEndTs`), cleared by every recompute. In-order live events cannot diverge, so hot keys skip too.
- New keyed state `giga-tile-large-ir-dirty`: null (fresh key, or a checkpoint from before the state existed) reads as **dirty**, so one recompute settles it — a divergence captured by an old checkpoint cannot outlive restore. An older binary restoring a newer checkpoint ignores the state and recomputes unconditionally, as before.
- Side effect: a delayed eviction timer can no longer rewind `lastLargeRecomputeAsOfTs` (and the decayed view) to an older horizon while clean.

## Validation

- New regression test: evictions inside one tail hop must not scan daily slots or rewrite the running IR; hop and expiry boundaries must recompute and decay; a delayed timer must not rewind the marker; a batch-overlap replay must sweep between boundaries and settle the flag. Fails before the fix (3 recomputes where 1 expected).
- Full GigaTile suites green on 2.12: aggregator (StreamProcessor 34, BugRegression 14, Integration 4) and flink (ProcessFunction 53, FirstSeenKeyGrace 19, PushFlinkIntegration 3).
- Includes a whitespace-only commit fixing the two pre-existing scalafmt violations so `flink.checkFormat` passes.

## Checklist

- [x] Added Unit Tests
- [x] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


nikhil-zlai:
Thanks for the thorough trace — both actionable items addressed in b08dd96f6:

- **Blocking compile break: fixed.** `SerdeGigaTileStore` now implements the two members (my impl search had excluded test files — that's also how `online.test` was missing from the validation list). All four `GigaTileStore` implementations are covered, and `online[2.12.18].test` (GigaTileCodecRoundTripTest 4/4) is now part of the validated suites alongside aggregator (14+34+4) and flink (19+53+3).
- **Dirty delayed-timer rewind: clamped.** The eviction-path recompute now runs at `math.max(timerTs, lastLargeRecomputeAsOfTs)`, closing the rewind entirely rather than only narrowing it. The regression test gained the case you called out: a dirty delayed timer must sweep, must not rewind the marker, and must not resurrect the expired batch tail (the old-horizon recompute would have brought the tail back at 5L).

On the per-event `putLargeIrDirty(true)` write during replays: agreed it's plausible noise next to the tile writes on the same path — leaving it unguarded per your "measure before bothering," happy to add the read-guard if replay throughput profiling says otherwise.

Acknowledged open test gaps (non-blocking): dirty persisting across a deferred-batch window, and a Flink-level null-reads-as-dirty restore test. Can add either here or as a follow-up if you want them in this PR.

PR #2044, opened 2026-07-20
Bound sensor retries to one partition interval; trim sensor resources

nikhil-zlai:
## Problem

Salesforce DEV: sensor Spark jobs stuck in Running for many hours, relaunched endlessly, saturated the shared EMR Serverless application and queued everything behind them ([Slack thread](https://zipline-2kh4520.slack.com/archives/C09JALZBE69/p1784544345213349)). Every sensor — regardless of grid — polled for 96 × 15min = 24h while holding ~8 vCPUs, so an hourly node could stack up to 24 concurrent day-long waiters per dependency.

## Fix (micro, `ExternalSourceSensorUtil` only)

1. **Retry budget = one partition interval.** The schedule fires a fresh sensor every interval, so a sensor that outlives its own interval is redundant with its successor. The budget now derives from the dependency's grid: daily keeps the historical 96 × 15min = 24h budget byte-identically; an hourly sensor exits after ~1h; fine grids shrink the poll interval, with ceiling division so coverage always reaches one full interval (9m grid → 2m × 5).
2. **Resource trim.** Sensors mostly sleep between polls while EMR Serverless bills the held vCPUs. Driver 4→2 cores (one query thread, one for JVM housekeeping), executor 4→2 cores, memory unchanged: ~8 → 4 vCPUs per waiting sensor.

The root-cause conf issue on the customer side (a `time_partitioned` dependency over a daily-partitioned table crash-looping on `Unparseable date`) is handled by fixing the conf; with this change such a sensor self-terminates within one interval instead of holding capacity for a day.

## Testing

- New `ExternalSourceSensorUtilTest` case pins the budgets: daily 96×15min, hourly 4×15min, 15-min grid 5×3min, 9-min grid 5×2min (≥ one interval).
- `ExternalSourceSensorIntegrationTest` resource-override expectations updated.
- Full `api.test.planner.*` suites green (195 tests).


nikhil-zlai:
Also trimmed the sensor resource overrides in this PR: driver 4→1 cores, executor 4→2 cores (memory unchanged). Sensors mostly sleep between polls while EMR Serverless bills the held vCPUs — ~8 vCPU → 3 vCPU per waiting sensor. The driver only does catalog/manifest reads; the single 2-core executor covers the rare full-scan fallback tier.

tchow-zlai (approved):

PR #2053, opened 2026-07-22
Fail KV metadata upload when the underlying put returns false

piyush-zlai:
## Summary
- `DynamoDBKVStoreImpl.multiPut` converts a per-item write failure (e.g. Dynamo's "Item size has exceeded the maximum allowed size" error) into `Success(false)` rather than a failed `Future`, so `Await.result` in `KVUploadNodeRunner` never threw.
- Root cause found while investigating why bringing `ranking.search.attempt__33` online in `#ext-depop` logged a DynamoDB error but the KV upload node runner still reported success.
- `doUploadJoinMetadata` and `doUploadModelTransforms` now check the returned boolean and throw a `RuntimeException` when the underlying put fails, so the job correctly fails instead of silently dropping the write.

## Test plan
- [x] Added regression tests asserting the runner throws when `kvStore.put` resolves to `false`, for both `JOIN_METADATA_UPLOAD` and `MODEL_TRANSFORMS_UPLOAD`.
- [x] Fixed the two pre-existing "handle ... successfully" tests, which turned out to rely on Mockito's default `false` return value rather than an explicit stub — they now stub `put` to return `true` and genuinely assert the happy path.
- [x] `./mill spark.test.testOnly "ai.chronon.spark.kv_store.KVUploadNodeRunnerTest"` — 9/9 passing.
- [x] `./mill spark.compile spark.test.compile` — clean.

## Follow-up (not in this PR)
Two related swallowed-failure risks in the DynamoDB bulk-load path (`bulkPut`) will be folded into the upcoming fix for the oversized-join-conf issue:
- the batch-table registry write after an `ImportTable` also discards its `multiPut` result
- `ImportTable` can report `COMPLETED` with per-item `errorCount() > 0`, which currently only gets an info log

## Summary by CodeRabbit

* **Bug Fixes**
  * Metadata uploads now detect unsuccessful KV store writes and report failures instead of treating them as successful.
  * Improved error handling for join metadata and model transform uploads, with clear failure reporting.


piyush-zlai:
Closing in favor of #2054 (bmagyar/join-metadata-upload), which independently includes the same KV-put-failure fix and also adds compression for oversized Join metadata. Folding the registry-write and per-item-error fixes for the bulk-load path into that branch as well, plus extending compression to ModelTransforms and switching the codec to operate on raw bytes instead of base64 text.

PR #2057, opened 2026-07-23
Surface branch and effective engine version as k8s labels on Flink jobs + wait for job manager deployment status error transitions

piyush-zlai:
## Summary
Covers a couple of items:
* Add support in the K8sFlinkSubmitter and the chain to plumb through additional labels (and the work to sanitize them for k8s). This allows us to pass labels like the branch, engine version etc.
* I noticed on the odd submission that jobs can prematurely be marked as failed due to jmDeploymentStatus being error. Turns out this can happen during one of issues in startup containers like init-jars which can be sorted on the init container restarts within k8s. So rather than prematurely marking as failed, we'll apply our grace period for the jmDeploymentStatus = error case as well. 

## Testing
Primarily the first piece (second shows up in a rare manner so hard to test outside uts). Confirmed that the labels are indeed being set / plumbed down - 
```
kubectl get pod flink-819dbe18-026f-443b-a1e1-d5a70ffbafcf-856647897b-b46p4 -n zipline-flink -o jsonpath='{.metadata.labels}' | jq
{
  "app": "flink-819dbe18-026f-443b-a1e1-d5a70ffbafcf",
  "branch": "piyush_flink-branch-label",
  "chronon/job_name": "aws_user_activities_label_v1__1",
  "component": "jobmanager",
  "job-type": "flink",
  "metadata-name": "aws.user_activities.label_v1__1__streaming",
  "pod-template-hash": "856647897b",
  "run-mode": "deploy",
  "type": "flink-native-kubernetes",
  "user": "piyush_zipline.ai",
  "zipline-team": "aws",
  "zipline-version": "1.18.0",
  "zipline_workflow_id": "a47314d1-fec9-4a09-83b0-2d7aa2be73f7"
...
}
```

## Checklist
- [X] Added Unit Tests
- [X] Covered by existing CI
- [X] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added support for forwarding submission labels to Flink jobs across AWS and Azure.
  * Applied sanitized pod/deployment labels for Kubernetes Flink submissions, including per-submission labels with precedence rules.
  * Added branch metadata as a built-in submission label.
* **Bug Fixes**
  * Improved handling of transient deployment errors: new deployments remain pending until timeout rather than failing immediately.
* **Tests**
  * Expanded unit tests for label propagation, sanitization, precedence, and deployment status timeout behavior.


tchow-zlai (changes requested):
Two correctness issues should be addressed before merge:

1. **[P1] Validate or sanitize label keys too** (`K8sFlinkSubmitter.scala:146`). `labels` is the generic `JobSubmitter.submit` input, so callers can supply keys that are valid for other backends but invalid for Kubernetes labels (for example `team name` or an over-63-character name). This code sanitizes only values and then puts every key into pod metadata, causing the entire FlinkDeployment admission to fail. Please validate/normalize (or explicitly reject) keys, including prefix/name length rules, and add an invalid-key test.

2. **[P1] Measure the ERROR grace period from the error transition, not CR creation** (`K8sFlinkSubmitter.scala:242-243`). Any deployment that has run longer than `DeploymentPendingTimeout` and later enters a transient `DEPLOYED/ERROR` state during an operator/Kubernetes restart is returned as `FAILED` on the first poll—the premature transition this change intends to prevent. The tests only exercise newly created CRs and miss this case. Please use a relevant status transition/reconciliation timestamp (or track first observation of `ERROR`) and cover an old CR that has only recently entered `ERROR`.

CI is green, and the label plumbing itself is covered well.

tchow-zlai (commented):
Do we need equivalent for the crucible submission?

tchow-zlai (approved):

PR #2058, opened 2026-07-23
fix(online): keep serving existing metadata when a TTLCache refresh fails

cceasy:
## Summary
`TTLCache`'s background refresh overwrote the cached value **unconditionally**. For the caches whose loader **returns** a `Failure` sentinel instead of throwing (`getGroupByServingInfo`, `buildJoinCodecCache`), a single transient backing-store blip during refresh would replace a previously-good entry with a `Failure` and **poison** it. Every subsequent request for a join using that GroupBy then failed until a refresh finally succeeded.

This change makes `TTLCache` **serve the last-known-good value when a refresh fails** (serve-stale), and adds per-key jitter so entries don't all expire and re-fetch on the same tick.

## Background / what we saw in our prod

On our HK prod fetcher, BigTable reads were intermittently hitting the client's per-attempt RPC deadline (`DEADLINE_EXCEEDED`, ~250 ms). During a serving-info refresh this produced a `Failure`, which got cached, and the impact fanned out well beyond the blip:

```mermaid
flowchart TD
    A["BigTable read times out<br/>(DEADLINE_EXCEEDED ~250ms)"] --> B["getGroupByServingInfo loader<br/>returns Failure(...)"]
    B --> C["TTLCache overwrites the good entry<br/>with the Failure (poisoning)"]
    C --> D["Every request for the join hits<br/>the cached Failure for the whole<br/>slow window, not just the one blip"]
    D --> E1["join.fetch.exception storm"]
    D --> E2["derivations can't resolve columns<br/>→ UNRESOLVED_COLUMN"]
    D --> E3["partial response fails Avro encode<br/>→ 'Logging failed due to ...'"]
```

Because the failure was **cached**, downstream retries within their request budget could not recover — they got the same instant failure until a later refresh landed under the deadline.

## Root cause

`TTLCache.asyncUpdateOnExpiry` did `cMap.put(i, Entry(wrappedCreator(i), now))` regardless of the result. Two idioms existed in the codebase:

- `getJoinConf` / `getModelTransformsConf` **throw** on failure — a throw never reaches the cache, so they were already safe.
- `getGroupByServingInfo` / `buildJoinCodecCache` **return** `Failure(...)` — so the failure was cached and poisoned the entry.

## Changes
1. Serve-existing on refresh failure — TTLCache now keeps the last-known-good entry when a background refresh fails instead of caching the failure;   and a cache.refresh_failure metric is emitted for alerting.                                                                                                  
2. Jittered expiry — a deterministic per-key refreshJitterRatio (default 0.1) shifts entries' expiry by up to 10% so they don't all refresh on the same tick and stampede the backing store.     

## Checklist
- [ ] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TTL cache refresh so failed or invalid refresh results no longer overwrite the last known-good value.
  * Ensured transient refresh issues (such as temporary lookup misses/timeouts) don’t “poison” previously cached successful entries.
  * Added deterministic per-key refresh jitter to spread cache refreshes more evenly over time.
* **Tests**
  * Added TTL cache test suite covering async refresh failure handling, recovery after successful refreshes, cold-start failure behavior, and validation of refresh jitter configuration.


cceasy:
CI check could be succeed on retry, as it succeed in my local.

piyush-zlai (commented):
comment on cold start behavior, overall the changes look good to me.

piyush-zlai (approved):

PR #2072, opened 2026-07-28
Redis implementation read optimizations

BenMagyar:
## Summary

Off the prior benchmarks, we're optimizing over the read path for some wide joins. Changes here are a lot of harness changes but:

- Update the benchmarks for some correctness issues around deduping over GroupBys
- Added additional timing for cache lookup, planning and response fanout

Improvements: 

- +3142% on benchmarks - The largest change was to swap to a planned path for `GET` and `ZRANGE` reqs.  We first construct a planned set of commands and then queue those onto one Redis cluster pipeline. This was _really_ beneficial. 
- +8.4% - We swapped from `ZRANGEWITHSCORES` to `ZRANGE` given the timestamp is already embedded in the value and the timestamp is the score we do not need to read it. 
- +18.2% - We stopped validating pools with PING reqs on every borrow, instead they get validated in the bg when idle. 

Other changes:
- Metrics were being repeatedly written with the same values per batch (this doesn't really change that much I think)
- JoinPart linear dedup - made the dedupe path on the JoinParts a little faster
- Not every temporal GroupBy has a topic attached so they do not all have written tiles - we only check now if there is an attached topic. 
 
## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added configurable Redis key hash-tag modes to support flexible data placement.
  * Added configurable Redis connection-pool idle limits with validation and clearer startup reporting.

* **Bug Fixes**
  * Improved Redis batch reads with unified pipelining, retries, and ASK redirect recovery.
  * Prevented unnecessary streaming reads for temporal data without a streaming source.
  * Improved cache hit/miss handling and request deduplication.
  * Preserved request ordering and response consistency during mixed reads.

### DIFF
--- a/online/src/main/scala/ai/chronon/online/fetcher/FetcherCache.scala
+++ b/online/src/main/scala/ai/chronon/online/fetcher/FetcherCache.scala
@@ -120,38 +120,33 @@ trait FetcherCache {
   def getCachedRequests(
       groupByRequestToKvRequest: Seq[(Fetcher.Request, Try[LambdaKvRequest])]): Map[GetRequest, CachedBatchResponse] = {
 
-    def empty = Map.empty[GetRequest, CachedBatchResponse]
+    val empty = Map.empty[GetRequest, CachedBatchResponse]
 
     if (!isCacheSizeConfigured) return empty
 
-    groupByRequestToKvRequest
-      .map {
+    val cachedRequests = Map.newBuilder[GetRequest, CachedBatchResponse]
+    groupByRequestToKvRequest.foreach {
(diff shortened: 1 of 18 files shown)


piyush-zlai (commented):
The redis side lgtm - I flagged a couple of aspects on the core fetcher that i'm not sure if they're needed / not. Would be safer if they're not crucial to revert for now to not trigger regressions while still keeping the redis optimizations

piyush-zlai (approved):

PR #2075, opened 2026-07-28
Fetcher join path optimizations

BenMagyar:
## Summary

Stacked on [Redis implementation read optimizations](https://github.com/zipline-ai/chronon/pull/2072).

Off the prior Redis read benchmarks, this pulls out the fetcher join path work separately. Changes here are a lot of harness changes but:

- Adds a full `Fetcher.fetchJoin` benchmark for batches of 50 candidates and 50 GroupBys, including the contextual and candidate request split
- Resolves join metadata once per join name for each fetch batch
- Skips the per-request external setup when a join has no external parts
- Handles contextual-only external parts inline instead of building and scheduling the external request fanout

Improvements:

- +1.3% at concurrency 16 for resolving join metadata once per batch (+6.2% serial)
- +10.7% for the empty-external path
- +1.8% for contextual-only inputs, from 1,016.46 to 1,034.68 batches/sec
- Executor work for contextual-only inputs drops from 9 to 5 tasks per batch (-44.4%), with peak queue depth dropping from 15 to 6

The percentages are isolated changes and should not be added together.

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


piyush-zlai (commented):
one test ask, otherwise lgtm

PR #2077, opened 2026-07-28
Reuse Redis cluster pipeline executors

BenMagyar:
## Summary

Stacked on [Redis implementation read optimizations](https://github.com/zipline-ai/chronon/pull/2072).

Jedis 5.1 creates and tears down a fixed thread pool for each Redis cluster pipeline sync. At the batch rates we are testing this creates thousands of short-lived reader threads each second. Changes here:

- Upgrades Jedis from 5.1.0 to 7.5.2
- Reuses a shared Redis pipeline executor instead of creating one per pipeline
- Defaults the shared pool to 12 threads and keeps `REDIS_PIPELINE_SYNC_THREADS=0` as the rollback to the Jedis-managed behavior
- Adds benchmark telemetry for pipeline executor activity, queue depth and task counts

Improvements from the longer isolated A/B on the fixed 4-CPU Redis fixture:

- 769.97 to 862.53 batches/sec (+12.02%)
- p95 latency drops 21.36%
- p99 latency drops 36.74%
- Replaces roughly 9,000 timed reader-thread creations with 12 persistent drain threads handling the same 9,000 tasks

The sizing sweep had 12 threads as the best point. 48 shared threads regressed 7.78%, so this uses a fixed default rather than deriving it from the client CPU count.

Both sides of the A/B used the same key layout, so the percentage isolates the executor change rather than claiming the absolute number as default-layout capacity.

## Checklist
- [x] Added Unit Tests
- [ ] Covered by existing CI
- [ ] Integration tested
- [ ] Documentation update


piyush-zlai (approved):
Small suggestion on sizing the pipeline tp..

